PR #10, opened 2016-10-04
Internet Explorer fixes

captn3m0:


ankitstarski:
:+1: LGTM

ankitstarski:
@captn3m0 you might want to remove `onclick` from this line [1], otherwise `openCheckout()` will fire twice.

 [1] https://github.com/razorpay/razorpay-woocommerce/blob/303644c84fc652d13329e57186f667a4a660ccea/razorpay-payments.php#L179

razorpay-payments.php
```diff
@@ -180,9 +180,13 @@ public function generate_razorpay_form($order_id)
     <button id="btn-razorpay-cancel" onclick="document.razorpayform.submit()">Cancel</button>
 </p>
 <script>
-    var setDisabled = function(id, state = true) {
```
ankitstarski: default state is `true` here and `false` after patch

captn3m0:
@ankitstarski The onclick was removed. change got stashed accidentally.

razorpay-payments.php
captn3m0: Oops. Will fix.

razorpay-payments.php
```diff
@@ -215,8 +219,20 @@ function openCheckout() {
       // Disable the pay button
       setDisabled('btn-razorpay');
```
ankitstarski: This may enable the `pay-razorpay` button

PR #11, opened 2016-10-14
New logo

mayankamencherla:
Switched the razorpay logo to the transparent one


mayankamencherla:
Just discovered that the razorpay logo was already transparent. There's something wrong with the way woocommerce is displaying it..... Closing the PR

PR #18, opened 2016-11-09
Webhook Support

mayankamencherla:


razorpay-payments.php
```diff
@@ -102,7 +108,14 @@ function init_form_fields()
...
                     )
-                )
+                ),
+                'enable_webhook' => array(
+                    'title' => __('Enable Webhook', 'razorpay'),
+                    'type' => 'checkbox',
+                    'description' => esc_url( admin_url('admin-post.php') ) . '?action=rzp_webhook',
+                    'label' => __('Enable Razorpay Webhook with the URL listed below.', 'razorpay'),
+                    'default' => 'yes'
+                ),
```
captn3m0: Add a `webhook_action` here that decides what we do with the order on payment success. There are multiple scenarios:

- Late Authorized (we get payment in authorized status, let the store owner decide)
- payment.failed (we can mark as failed) (very tricky, let us leave)
- Actual authorized, but post to woocommerce failed (we let store decide)

Store decisions:

1. Mark the payment as successful, if there is stock/inventory
2. Inform the store owner?

razorpay-payments.php
```diff
@@ -8,10 +8,14 @@
 Author URI: https://razorpay.com
 */
 
+require_once __DIR__.'/includes/razorpay-webhook.php';
+
 require_once __DIR__.'/razorpay-sdk/Razorpay.php';
 use Razorpay\Api\Api;
 
 add_action('plugins_loaded', 'woocommerce_razorpay_init', 0);
+add_action('admin_post_nopriv_rzp_webhook', 'razorpay_webhook_init'); // second - admin_post_nopriv
```
captn3m0: do we need both of these?

razorpay-payments.php
```diff
@@ -217,85 +230,12 @@ function get_order_creation_data($order_id)
         **/
         function generate_order_form($redirect_url, $json, $order_id)
         {
+            $checkout_html = file_get_contents(__DIR__.'/js/checkout.phtml');
+            $keys = array("#liveurl#", "#json#", "#redirect_url#");
```
captn3m0: will need to be rebased against master

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,66 @@
...
+
+    function auto_capture_webhook()
+    {
+        $razorpay = new WC_Razorpay();
+
+        $post = file_get_contents('php://input');
+
+        $data = json_decode($post, true);
+
+        if ($razorpay->enable_webhook == 'yes' && !empty($data['event']))
```
captn3m0: make yes into constant, or use a boolean instead if possible

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,66 @@
...
+        {   
+            // if payment.authorized webhook is enabled, we will update woocommerce about captured payments
+            if ($data['event'] == "payment.authorized")
```
captn3m0: Check if order is paid at WooCommerce end before running this code. For most payments, that will be the case.

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,66 @@
...
+            {
+                global $woocommerce;
+                $order_id = $data['payload']['payment']['entity']['notes']['woocommerce_order_id'];
+                $order = new WC_Order($order_id);
+                
+                $key_id = $razorpay->key_id;
+                $key_secret = $razorpay->key_secret;
```
captn3m0: 1. get the payment id from the order
2. update the order (mark as successful)
3. do the same steps as we do in normal payment success flow

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,66 @@
...
+
+                $razorpay_payment_id = $data['payload']['payment']['entity']['id'];
+
+                $api = new Api($key_id, $key_secret);
+
+                $payment = $api->payment->fetch($razorpay_payment_id);
+
+                if ($payment['status'] == 'captured')
+                {
```
captn3m0: try re-using the existing code here?

razorpay-payments.php
captn3m0: We need to add instructions about how to configure webhooks on the website. Something like:

>Use https://www.woocommerce.com/adlkasd/asfdsaf.php as the Webhook URL [here](https://dashboard.razorpay.com/#/app/webhooks) and enable the `payment.authorized` event _only_.

We will also need Signature verification in place (add another input for webhook secret (minlength=8))

razorpay-payments.php
mayankamencherla: @captn3m0 if you want to have both logged in and not logged in users submitting, you have to add both actions!

https://codex.wordpress.org/Plugin_API/Action_Reference/admin_post_(action)

razorpay-payments.php
mayankamencherla: @captn3m0 Update or Inform store are my 2 options.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 I believe wordpress needs a yes or no here.

razorpay-payments.php
mayankamencherla: @captn3m0 Will do towards the end once we finalize the functionality.

includes/razorpay-webhook.php
mayankamencherla: Performing this check after we get the $order object.  

```
if ($order->needs_payment() === false)
{
         return;
}
```

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Don't get it? Don't we already mark the order as successful here?

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Will have to rebase against  master first.

razorpay-payments.php
```diff
@@ -472,3 +425,8 @@ function woocommerce_add_razorpay_gateway($methods)
 
     add_filter('woocommerce_payment_gateways', 'woocommerce_add_razorpay_gateway' );
 }
+
+function razorpay_webhook_init()
+{
+    new RZP_Webhook();
+}
```
captn3m0: 1. newline at the end. 
2. I don't like doing everything in the constructor here

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+require_once __DIR__.'/../razorpay-payments.php';
```
captn3m0: Found a couple of nice suggestions on how to do webhooks in wordpress cleanly:

- https://wordpress.stackexchange.com/questions/178090/best-way-for-plugin-to-accept-posts
- https://wordpress.stackexchange.com/questions/126572/trying-to-avoid-including-wp-load-php/126582#126582

razorpay-payments.php
captn3m0: Hard to review, because it does HTML very differently

razorpay-payments.php
captn3m0: We don't have logged-in-users submitting in this case, it is the razorpay servers

includes/razorpay-webhook.php
captn3m0: I wrote (1) based on order.paid, so ignore. However, do try to re-use the success/failure code, instead of copying here

includes/razorpay-webhook.php
captn3m0: We should do an early `die` as well, so other plugins and wordpress does not interfere with our webhooks.

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,78 @@
...
+
+require_once __DIR__.'/../razorpay-sdk/Razorpay.php';
+use Razorpay\Api\Api;
+
+class RZP_Webhook 
+{
+    protected $razorpay; 
+
+    protected $keyId;
+    protected $keySecret;
```
captn3m0: this looks duplication. What about creating a WC_Razorpay_Webhook instance in our primary class and using that?

razorpay-payments.php
mayankamencherla: @captn3m0 Merged

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 As discussed just pulling out the $api from getRazorpayApiInstance()

razorpay-payments.php
mayankamencherla: @captn3m0 Using only $api in the constructor. IE. $api.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 I tried adding an endpoint. Although, I think using admin post is the way to go, I remember not finding results with the approach above.

js/checkout.phtml
```diff
@@ -0,0 +1,71 @@
+<script src="#liveurl#"></script>
```
captn3m0: This file is not needed

razorpay-payments.php
```diff
@@ -8,7 +8,11 @@
 Author URI: https://razorpay.com
 */
 
-if ( ! defined( 'ABSPATH' ) ) {
+require_once __DIR__.'/includes/razorpay-webhook.php';
+require_once __DIR__.'/razorpay-sdk/Razorpay.php';
+
+if ( ! defined( 'ABSPATH' ) )
```
captn3m0: The ABSPATH check should be the first thing in the file

razorpay-payments.php
```diff
@@ -44,6 +51,10 @@ public function __construct()
             $this->key_secret = $this->settings['key_secret'];
             $this->payment_action = $this->settings['payment_action'];
 
+            $this->enable_webhook = $this->settings['enable_webhook'];
+
+            $this->liveurl = 'https://checkout.razorpay.com/v1/checkout.js';
```
captn3m0: not needed

razorpay-payments.php
```diff
@@ -44,6 +51,10 @@ public function __construct()
             $this->key_secret = $this->settings['key_secret'];
             $this->payment_action = $this->settings['payment_action'];
 
+            $this->enable_webhook = $this->settings['enable_webhook'];
```
captn3m0: This broke the last time we did a release with `payment_action`. We had to ask users to visit settings and submit it again so it entered the database. Have a default value here, so it doesn't give errors.

razorpay-payments.php
```diff
@@ -109,7 +120,29 @@ function init_form_fields()
...
+                    'label' => __('Enable Razorpay Webhook <a href="https://dashboard.razorpay.com/#/app/webhooks">here</a> with the URL listed below.', 'razorpay'),
+                    'default' => 'yes'
+                ),
+                'webhook_action' => array(
+                    'title' => __('Webhook Action', 'razorpay'),
+                    'type' => 'select',
+                    'description' =>  __('Would you like our webhooks to update your store, or inform you so that you can update it yourself?', 'razorpay'),
+                    'default' => 'update',
+                    'options' => array(
+                        'update' => 'Update Store',
```
captn3m0: This is very confusing. Needs helper text about what each option does.

razorpay-payments.php
```diff
@@ -109,7 +120,29 @@ function init_form_fields()
...
+                        'inform'   => 'Inform store'
+                    )
+                ),
+                'webhook_secret' => array(
+                    'title' => __('Webhook Secret', 'razorpay'),
+                    'type' => 'text',
+                    'description' => __('Webhook secret is used for webhook signature verification. This has to match the one added <a href="https://dashboard.razorpay.com/#/app/webhooks">here</a>', 'razorpay')
```
captn3m0: There should be some helper guide about how to integrate the webhook properly. Perhaps link them to a page on the repo wiki on github?

razorpay-payments.php
```diff
@@ -371,86 +404,6 @@ private function enqueueCheckoutScripts($data)
         function generateOrderForm($redirectUrl, $data)
         {
             $this->enqueueCheckoutScripts($data);
-
-            return <<<EOT
-<form name='razorpayform' action="$redirectUrl" method="POST">
```
captn3m0: Why are we removing this?

razorpay-payments.php
mayankamencherla: @captn3m0 the enable_webhook setting has a default in the settings set to `yes`

includes/razorpay-webhook.php
captn3m0: We want to give a absolute URL to the user that they can enter in the Razorpay dashboard in the settings page.

razorpay-payments.php
captn3m0: We had that for payment_capture as well, and it broke. Can you try deleting the entry from your database and checking if it raises a notice?

razorpay-payments.php
mayankamencherla: @captn3m0 My bad. Added it.

razorpay-payments.php
mayankamencherla: @captn3m0 Will add it after creating the wiki

razorpay-payments.php
```diff
@@ -109,7 +125,29 @@ function init_form_fields()
...
+                'enable_webhook' => array(
+                    'title' => __('Enable Webhook', 'razorpay'),
+                    'type' => 'checkbox',
+                    'description' => esc_url( admin_url('admin-post.php') ) . "?action=rzp_wc_webhook <br><br>Instructions and guide to <a href='https://github.com/razorpay/razorpay-woocommerce/wiki/Razorpay-Woocommerce-Webhooks'>Razorpay webhooks</a>",
+                    'label' => __('Enable Razorpay Webhook <a href="https://dashboard.razorpay.com/#/app/webhooks">here</a> with the URL listed below.', 'razorpay'),
+                    'default' => 'yes'
+                ),
+                'webhook_action' => array(
+                    'title' => __('Webhook Action', 'razorpay'),
+                    'type' => 'select',
```
captn3m0: I don't see webhook_action being used anywhere

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,61 @@
...
+                }
+
+                $razorpayPaymentId = $data['payload']['payment']['entity']['id'];
+
+                $payment = $this->api->payment->fetch($razorpayPaymentId);
+
+                $success = false;
+                $errorMessage = 'The payment has failed.';
+
+                if ($payment['status'] === 'captured')
```
captn3m0: Shouldn't we be capturing the payment in case it was authorized?

razorpay-payments.php
```diff
@@ -44,6 +50,17 @@ public function __construct()
...
+            {
+                $this->enable_webhook = $this->settings['enable_webhook'];
+            }
+            else
+            {
+                $this->enable_webhook = 'yes';
+            }
+
+            $this->webhook_secret = $this->settings['webhook_secret'];
+
```
captn3m0: default value needed.

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,97 @@
...
+    protected $razorpay;
+    protected $api;
+
+    function __construct()
+    {
+        $this->razorpay = new WC_Razorpay();
+
+        $this->api = $this->razorpay->getRazorpayApiInstance();
+
+        $this->auto_capture_webhook();
```
captn3m0: have a public process function instead, and call the appropriate method depending on the event (payment_authorized in our case). Makes it pluggable for subscriptions

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,97 @@
...
+        if ($this->razorpay->enable_webhook === 'yes' && empty($data['event']) === false)
+        {
+            if ((isset($_SERVER['HTTP_X_RAZORPAY_SIGNATURE']) === true))
+            {
+                $razorpayWebhookSecret = $this->razorpay->webhook_secret;
+
+                //
+                // If the webhook secret isn't set on wordpress, return
+                //
+                if (isset($razorpayWebhookSecret) === false)
```
captn3m0: isset will be true for empty strings

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,97 @@
...
+                    return;
+                }
+            }
+
+            // if payment.authorized webhook is enabled, we will update woocommerce about captured payments
+            // We have to complete the payment only if the order needs payment
+            if ($data['event'] === "payment.authorized")
+            {
+                global $woocommerce;
+                $orderId = $data['payload']['payment']['entity']['notes']['woocommerce_order_id'];
```
captn3m0: Use $order->receipt instead. I'm not very sure if we have order entity there.

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,97 @@
...
+                $order = new WC_Order($orderId);
+
+                // Move this to the base class as well
+                if ($order->needs_payment() === false)
```
captn3m0: We need the same check on the other side as well, right?

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 As discussed let us leave as is

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,106 @@
...
+        $this->razorpay = new WC_Razorpay();
+
+        $this->api = $this->razorpay->getRazorpayApiInstance();
+    }
+
+    public function process()
+    {
+        $post = file_get_contents('php://input');
+
+        $data = json_decode($post, true);
```
captn3m0: test it against an empty post body

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,106 @@
...
+            $success = true;
+        }
+        else if (($payment['status'] === 'authorized') and
+                 ($this->razorpay->payment_action === 'capture'))
+        {
+            //
+            // If the payment is only authorized, we capture it
+            // If the merchant has enabled auto capture
+            //
+            $payment->capture(array('amount' => $amount));
```
captn3m0: we can't capture with the amount from the webhook, we need to capture with the order entity in WC

razorpay-payments.php
```diff
@@ -109,7 +133,19 @@ function init_form_fields()
...
-                )
+                ),
+                'enable_webhook' => array(
+                    'title' => __('Enable Webhook', 'razorpay'),
+                    'type' => 'checkbox',
+                    'description' => esc_url( admin_url('admin-post.php') ) . "?action=rzp_wc_webhook <br><br>Instructions and guide to <a href='https://github.com/razorpay/razorpay-woocommerce/wiki/Razorpay-Woocommerce-Webhooks'>Razorpay webhooks</a>",
+                    'label' => __('Enable Razorpay Webhook <a href="https://dashboard.razorpay.com/#/app/webhooks">here</a> with the URL listed below.', 'razorpay'),
+                    'default' => 'yes'
+                ),
+                'webhook_secret' => array(
```
captn3m0: Let us make it mandatory if the webhook is enabled.

razorpay-payments.php
```diff
@@ -527,14 +572,19 @@ function check_razorpay_response()
...
                 $this->handleErrorCase($order);
             }
 
             $this->updateOrder($order, $success, $error, $razorpayPaymentId);
 
+            $this->redirectUser($order);
+        }
+
+        protected function redirectUser($order)
+        {
```
captn3m0: in the early return case, $this->msg is not set

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Made getOrderAmountAsInteger public

razorpay-payments.php
mayankamencherla: @captn3m0 You're right. My bad. Added the code below:

```
if ($order->needs_payment() === false)
            {
                $id = $this->getOrderId($order);

                $razorpayPaymentId = get_post_meta($id, '_transaction_id');

                $this->updateOrder($order, true, null, $razorpayPaymentId);

                $this->redirectUser($order);
            }
```

mayankamencherla: @captn3m0 This should be good to go unless the `$razorpayPaymentId` isn't set. Do we handle that case too? If the order is already marked paid by the webhook, then surely the $razorpayPaymentId should be saved right?

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Handled. Tested. 
razorpay-webhook.php - process function. 
We check for whether data has event set.

razorpay-payments.php
mayankamencherla: @captn3m0 How about generating a random timestamp based string and storing that as a default value?

We can add instructions on how to set up the secret, ie, by copying the string we generated into the dashboard, or by creating another one and storing it in both of the locations.

razorpay-payments.php
captn3m0: Do we need to update the order at all in that case?

razorpay-sdk/src/Utility.php
```diff
@@ -17,14 +17,21 @@ public function verifyPaymentSignature($attributes)
...
     {
-        return self::verifySignature($payload, $expectedSignature);
+        return self::verifySignature($payload, $expectedSignature, $webhookSecret);
     }
 
-    public function verifySignature($payload, $expectedSignature)
+    public function verifySignature($payload, $expectedSignature, $webhookSecret = '')
     {
-        $actualSignature = hash_hmac(self::SHA256, $payload, Api::getSecret());
+        if (isset($webhookSecret) === true)
```
captn3m0: This will always return true.

includes/razorpay-webhook.php
captn3m0: This will break if the body is empty or non-valid JSON

includes/razorpay-webhook.php
```diff
@@ -0,0 +1,106 @@
...
+                    return;
+                }
+
+                try
+                {
+                    $this->api->utility->verifyWebhookSignature($post,
+                                                                $_SERVER['HTTP_X_RAZORPAY_SIGNATURE'],
+                                                                $razorpayWebhookSecret);
+                }
+                catch (Exception $e)
```
captn3m0: Don't catch all exceptions and we need to log here.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 If it is empty, it will return null. But yeah, if it isn't a valid json, it'll break. 

Either ways, `json_last_error` will be anything but 0 if json_decode failed for whatever reason. So I'll add that.

razorpay-payments.php
mayankamencherla: @captn3m0 You don't. 
I moved 
`$this->add_notice($this->msg['message'], $this->msg['class']);`
to 
`public function updateOrder(& $order, $success, $errorMessage, $razorpayPaymentId)`

razorpay-payments.php
captn3m0: I like this idea, but there are complications around when you generate this secret. You don't want to do it every time they open or submit the settings page.

includes/razorpay-webhook.php
captn3m0: Valid json without `event` key will also break

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 write_log(['message' => $e->getMessage()]);

razorpay-sdk/src/Utility.php
mayankamencherla: @captn3m0 Changed this to empty() === false
Also, default is set to `''`.

razorpay-payments.php
captn3m0: Won't the order be updated twice in that case? The payment note gets added twice, right?

razorpay-payments.php
mayankamencherla: @captn3m0 Right, I had thought of the case as well. What then, we'll just leave it to null then?

razorpay-payments.php
mayankamencherla: @captn3m0 Changed this.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 How? We're doing an empty check on line 32!

razorpay-sdk/src/Utility.php
captn3m0: @mayankamencherla I'd pointed this out, and you'd mentioned you fixed it :-1:

PR #23, opened 2016-12-16
Added dependencies and curl installation to readme

mayankamencherla:


captn3m0:
@mayankamencherla please amend the commit and drop the php-curl version.

PR #24, opened 2016-12-21
[session] Improves session handling

captn3m0:
- Namespaces the session that holds razorpay payment ID by the
  WC-order-id
- Also, do not recreate the order ID every time.

### DIFF
--- a/razorpay-payments.php
+++ b/razorpay-payments.php
@@ -153,21 +153,14 @@ function woocommerce_razorpay_init()
 
             $sessionKey = $this->getSessionKey($orderId);
 
-            $verify= true;
-
             try
             {
-                if ($woocommerce->session->get($sessionKey))
-                {
-                    $razorpayOrderId = $woocommerce->session->get($sessionKey);
-                    $verify = $this->verifyOrderAmount($razorpayOrderId, $orderId);
-                }
-                if ($verify === false)
-                {
-                    $razorpayOrderId = $this->createRazorpayOrderId(
-                        $orderId, $sessionKey);
-                }
-                else
+                $razorpayOrderId = $woocommerce->session->get($sessionKey);
+
+                // If we don't have an Order
+                // or the if the order is present in session but doesn't match what we have saved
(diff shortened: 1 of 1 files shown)


razorpay-payments.php
```diff
@@ -133,53 +134,71 @@ function receipt_page($order)
...
-                  ),
-                  'notes'           => array(
-                    'woocommerce_order_id' => $order_id
-                  ),
-                  'order_id'        => $razorpay_order['id']
-                );
-                $json = json_encode($razorpay_args);
-                $html = $this->generate_order_form($redirect_url, $json);
-                return $html;
+                // TODO: Verify that the order amount matches
```
captn3m0: @mayankamencherla We need a verify check that the razorpay order amount matches the order amount here.

razorpay-payments.php
```diff
@@ -133,53 +134,116 @@ function receipt_page($order)
...
-                );
-                $json = json_encode($razorpay_args);
-                $html = $this->generate_order_form($redirect_url, $json);
-                return $html;
+                if ($woocommerce->session->get($sessionKey))
+                {
+                    $razorpayOrderId = $woocommerce->session->get($sessionKey);
+                    $verify = $this->verifyOrderAmount($razorpayOrderId, $orderId);
+                }
+                else if ($verify === false)
```
captn3m0: @mayankamencherla The `else if` block will never be called, because verify is set to false _inside the if_.

razorpay-payments.php
```diff
@@ -133,53 +134,116 @@ function receipt_page($order)
...
+            $html = $this->generate_order_form($redirect_url,$json,$orderId);
+
+            return $html;
+        }
+
+        protected function createRazorpayOrderId($orderId)
+        {
+            // Calls the helper function to create order data
+            $data = $this->get_order_creation_data($orderId);
+            $razorpay_order = $api->order->create($data);
```
captn3m0: $api is not defined here.

PR #33, opened 2017-03-14
Discount coupon fix

mayankamencherla:
Fixed amount mismatch when discount coupon is applied

### DIFF
--- a/razorpay-payments.php
+++ b/razorpay-payments.php
@@ -171,8 +171,6 @@ function woocommerce_razorpay_init()
                 echo "RAZORPAY ERROR: Api could not be reached";
             }
 
-            die($order->order_total);
-
             $razorpay_args = array(
               'key'         => $this->key_id,
               'name'        => get_bloginfo('name'),
@@ -224,7 +222,7 @@ function woocommerce_razorpay_init()
 
             $razorpayOrderArgs = array(
                 'id'        => $razorpayOrderId,
-                'amount'    => (int) $order->order_total*100,
+                'amount'    => $order->order_total*100,
                 'currency'  => get_woocommerce_currency(),
                 'receipt'   => (string) $orderId,
             );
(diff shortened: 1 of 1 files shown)


razorpay-payments.php
```diff
@@ -222,7 +222,7 @@ protected function verifyOrderAmount($razorpayOrderId, $orderId)
 
             $razorpayOrderArgs = array(
                 'id'        => $razorpayOrderId,
-                'amount'    => (int) $order->order_total*100,
+                'amount'    => $order->order_total*100,
```
captn3m0: why was this changed?

mayankamencherla: @captn3m0 It was rounding 199 with 20% discount to 15919 instead of 15920

captn3m0: Dropping the int also means we may be passing float to our systems lower down which may cause issues, no? Did some tests:

```>>> $a  = 159.2;
=> 159.2
>>> (int) $a * 100;
=> 15900
>>> (int) ($a * 100);
=> 15919
>>> 159.2 * 100
=> 15920.0
>>> (int)(159.2 * 100);
=> 15919
>>> (159.2 * 100);
=> 15920.0
>>> floor($a * 100);
=> 15919.0
>>> round($a * 100);
=> 15920.0
>>> round($a * 100);
=> 15920.0
>>> intval($a * 100);
=> 15919
>>> ceil($a * 100);
=> 15920.0
>>> round($a * 100);
=> 15920.0
>>> (int) round ($a*100);
=> 15920
```

The typecast + round approach looks cleaner to me. /cc @73sl4

mayankamencherla: @captn3m0 Cool. Agreed. Pushed the code.

captn3m0:
Please test locally and do a squash merge.

PR #34, opened 2017-03-15
Display currency master

mayankamencherla:
- [x]  Added display currency to main woocommerce flow.
- [x] Saving display amount in sessions to avoid calling google finance api over and over again.


captn3m0:
Closing this

PR #35, opened 2017-05-26
WooCommerce 3.0 update

captn3m0:
Fixed some deprecations. Trying to keep the changes backwards compatible, so that we don't have to maintain 2 branches


captn3m0:
Tested against both WC2.6 and WC3.0

Will be merging soon, if no one has any concerns. /cc @riagarg5 @mayankamencherla

PR #40, opened 2017-06-07
Subscriptions

mayankamencherla:
* Create plan
* Create / fetch customer 
* Create subscription
* Create checkout args - orderId null if subscription, and vice versa. 

TODO: 
- [x] Throw error if more than one product in cart along with subscription
- [ ] Plan creation needs to be moved to product creation/edits instead of during payment process
- [x] All API calls to have a try catch block?
- [x] https://github.com/razorpay/razorpay-woocommerce/pull/40#discussion_r124869668


razorpay-payments.php
```diff
@@ -210,26 +229,146 @@ public function generate_razorpay_form($orderId)
...
             $html = $this->generateOrderForm($redirectUrl, $checkoutArgs);
 
             return $html;
         }
 
+        protected function getPlanArguments($product)
+        {
+            $productId = $product['product_id'];
+
+            $period       = WC_Subscriptions_Product::get_period($productId);
```
captn3m0: I'd recommend getting the product and calling all these over the product (these take both id and object as the argument). Would likely help us lower the DB calls

razorpay-payments.php
```diff
@@ -210,26 +229,146 @@ public function generate_razorpay_form($orderId)
...
+
+            return $periodMap[$period];
+        }
+
+        protected function getCustomerId($order)
+        {
+            $api = $this->getRazorpayApiInstance();
+
+            $data = array(
+                'name'          => $order->billing_first_name." ".$order->billing_last_name,
```
captn3m0: not compatible with 3.0

See https://github.com/razorpay/razorpay-woocommerce/pull/35

razorpay-payments.php
```diff
@@ -268,6 +406,18 @@ protected function getCheckoutArguments($order, $razorpayOrderId)
                 );
             }
 
+            // order_id and subscription_id both cannot be set at the same time
+            if ($razorpayOrderId)
+            {
+                $args['order_id'] = $razorpayOrderId;
+            }
+
+            if (empty($subscriptionId) === false)
```
captn3m0: splitting getCheckoutArgs to 2 classes would help here as well.

razorpay-sdk/src/Subscription.php
```diff
@@ -0,0 +1,21 @@
...
+    public function fetch($id)
+    {
+        return parent::fetch($id);
+    }
+
+    public function all($options = array())
+    {
+        return parent::all($options);
+    }
+}
```
captn3m0: leave newline in this and other files

razorpay-payments.php
```diff
@@ -52,6 +52,20 @@ public function __construct()
...
+                'subscriptions',
+                'subscription_cancellation',
+                'subscription_suspension',
+                'subscription_reactivation',
+                'subscription_amount_changes',
+                'subscription_date_changes',
+                'subscription_payment_method_change'
+            );
+
+            add_action('woocommerce_scheduled_subscription_payment', array($this, 'process_subscription'));
```
mayankamencherla: @captn3m0 Using it for now. Will remove it once we get the webhook code working.

razorpay-payments.php
```diff
@@ -210,26 +229,146 @@ public function generate_razorpay_form($orderId)
...
             $order = new WC_Order($orderId);
 
+            $subscriptionId = null;
+
+            if (wcs_order_contains_subscription($order) === true)
+            {
+                $products = $order->get_items();
+
+                // For each product, creating a new plan if not created already
+                foreach ($products as $key => $product)
```
mayankamencherla: @captn3m0 Added it in the TODO section.

razorpay-payments.php
```diff
@@ -210,26 +229,146 @@ public function generate_razorpay_form($orderId)
...
+                {
+                    $productId = $product['product_id'];
+
+                    // Check if product is a subscription product
+                    if (WC_Subscriptions_Product::is_subscription($productId) === true)
+                    {
+                        $metadata = get_post_meta($productId);
+
+                        // Creating a new plan only if plan isn't created already
+                        if (empty($metadata['plan_id']) === true)
```
mayankamencherla: @captn3m0 Move it to a constant?

razorpay-payments.php
```diff
@@ -210,26 +229,146 @@ public function generate_razorpay_form($orderId)
...
+                        {
+                            $api = $this->getRazorpayApiInstance();
+
+                            $planArgs = $this->getPlanArguments($product);
+
+                            $plan = $api->plan->create($planArgs);
```
mayankamencherla: @captn3m0 Added it to the TODO.

razorpay-payments.php
```diff
@@ -210,26 +229,146 @@ public function generate_razorpay_form($orderId)
...
+                                'item' => array(
+                                    'amount'   => (int) round($signUpFee * 100, 2),
+                                    'currency' => get_woocommerce_currency(),
+                                    'name'     => $product['name']
+                                )
+                            )
+                        );
+                    }
+
+                    $api = $this->getRazorpayApiInstance();
```
mayankamencherla: @captn3m0 Initializing $this->api in the constructor

razorpay-payments.php
```diff
@@ -210,26 +229,146 @@ public function generate_razorpay_form($orderId)
...
+                }
+            }
+
             $redirectUrl = get_site_url() . '/?wc-api=' . get_class($this);
 
             $razorpayOrderId = $this->createOrGetRazorpayOrderId($orderId);
 
-            if ($razorpayOrderId === null)
+            if (($razorpayOrderId === null) and
+                (wcs_order_contains_subscription($orderId) === false))
```
mayankamencherla: @captn3m0 No, We need to have this check even otherwise. 

$razorpayOrderId can be null when the order has a subscription.

razorpay-payments.php
```diff
@@ -210,26 +229,146 @@ public function generate_razorpay_form($orderId)
...
             {
                 return 'RAZORPAY ERROR: Api could not be reached';
             }
 
-            $checkoutArgs = $this->getCheckoutArguments($order, $razorpayOrderId);
+            $checkoutArgs = $this->getCheckoutArguments($order, $razorpayOrderId, $subscriptionId);
```
mayankamencherla: @captn3m0 Not a fan of this. Lot of code will be reused. Maybe I can just move the subscription bit over to the subscription file, as a small helper method? 

Let us discuss this once in person.

razorpay-payments.php
mayankamencherla: @captn3m0 See above. Let us discuss this feature once.

razorpay-payments.php
mayankamencherla: @captn3m0 Using $product isn't working like it should. Have to debug. Will add it to the TOOD.

razorpay-payments.php
```diff
@@ -539,6 +687,12 @@ function check_razorpay_response()
             exit;
         }
 
+        protected function process_subscription($subscription_id)
+        {
+            // This method is used to process the subscription's recurring payment
+            sd('1');
```
mayankamencherla: @captn3m0 My bad. Removed.

includes/razorpay-webhook.php
```diff
@@ -54,28 +54,45 @@ public function process()
...
-        global $woocommerce;
-
         //
         // Order entity should be sent as part of the webhook payload
         //
         $orderId = $data['payload']['payment']['entity']['notes']['woocommerce_order_id'];
 
+        $paymentId = $data['payload']['payment']['entity']['id'];
+
+        if (isset($data['payload']['payment']['subscription_id']) === true)
```
captn3m0: needs entity

includes/razorpay-webhook.php
```diff
@@ -110,4 +127,114 @@ protected function paymentAuthorized($data)
...
+        //
+        if ($success === false)
+        {
+            return $this->processSubscriptionFailed($orderId);
+        }
+
+        // This method is used to process the subscription's recurring payment
+        $wcSubscription = wcs_get_subscriptions_for_order($orderId);
+
+        $subscriptionId = get_post_meta($orderId, 'razorpay_subscription_id')[0];
```
captn3m0: use constant

includes/razorpay-webhook.php
```diff
@@ -110,4 +127,114 @@ protected function paymentAuthorized($data)
...
+     */
+    protected function processSubscriptionSuccess($wcSubscription, $subscription, $paymentId)
+    {
+        $wcSubscriptionId = array_keys($wcSubscription)[0];
+
+        // We will only process one subscription per order
+        $wcSubscription = array_values($wcSubscription)[0];
+
+        // The subscription is completely paid for
+        if ($wcSubscription->get_completed_payment_count() === $subscription->total_count)
```
captn3m0: save the output of get_completed_payment_count() in a variable. [WC seems to be caching it](https://github.com/wp-premium/woocommerce-subscriptions/blob/master/includes/class-wc-subscription.php#L712), but I'd rather still not call it twice.

includes/razorpay-webhook.php
```diff
@@ -110,4 +127,114 @@ protected function paymentAuthorized($data)
...
+        //
+        // If success is false, automatically process subscription failure
+        //
+        if ($success === false)
+        {
+            return $this->processSubscriptionFailed($orderId);
+        }
+
+        // This method is used to process the subscription's recurring payment
+        $wcSubscription = wcs_get_subscriptions_for_order($orderId);
```
captn3m0: Do array_values[0] manipulations here instead of inside processSubscriptionSuccess

razorpay-payments.php
```diff
@@ -76,6 +77,20 @@ public function __construct()
...
             $this->msg['class'] = '';
 
+            // Supports woocommerce subscriptions
+            $this->supports = array(
+                'products',
+                'subscriptions',
+                'subscription_cancellation',
+                'subscription_suspension',
+                'subscription_reactivation',
+                'subscription_amount_changes',
```
captn3m0: 1. This was already defined on line 71
2. We don't support all of these. Just stick to `subscriptions` for now. I'm not seeing any code for anything else (cancellation/suspension etc)

razorpay-payments.php
```diff
@@ -180,11 +195,16 @@ function receipt_page($order)
...
 
-        protected function getSessionKey($orderId)
+        protected function getOrderSessionKey($orderId)
         {
             return "razorpay_order_id.$orderId";
         }
 
+        protected function getSubscriptionSessionKey($orderId)
+        {
+            return "razorpay_subscription_id".$orderId;
```
captn3m0: nit: space around periods. Alternatively, just use string interpolation

razorpay-payments.php
```diff
@@ -282,23 +312,43 @@ protected function getCheckoutArguments($order, $razorpayOrderId)
...
               ),
-              'order_id'     => $razorpayOrderId,
               'callback_url' => $callbackUrl,
+              'prefill' => $this->getPreFillArguments($order)
             );
 
             $args['amount'] = $this->getOrderAmountAsInteger($order);
 
+            // order_id and subscription_id both cannot be set at the same time
+            if (empty($razorpayOrderId) === false)
```
captn3m0: Don't use empty for such things. Use null and set default value in method signature as null. Using empty causes type confusion.

razorpay-payments.php
```diff
@@ -593,20 +643,31 @@ protected function verifySignature($orderId)
...
             $attributes = array(
                 'razorpay_payment_id' => $_POST['razorpay_payment_id'],
-                'razorpay_order_id'   => $woocommerce->session->get($sessionKey),
                 'razorpay_signature'  => $_POST['razorpay_signature'],
             );
 
+            if (wcs_order_contains_subscription($orderId) === true)
+            {
+                $sessionKey = $this->getSubscriptionSessionKey($orderId);
+                $attributes['razorpay_subscription_id'] = $woocommerce->session->get($sessionKey);
```
captn3m0: make constant

razorpay-sdk/src/Utility.php
```diff
@@ -24,7 +34,7 @@ public function verifyWebhookSignature($payload, $expectedSignature, $webhookSec
 
     public function verifySignature($payload, $expectedSignature, $webhookSecret = '')
     {
-        if (isset($webhookSecret) === true)
+        if (empty($webhookSecret) === false)
```
captn3m0: Shouldn't this already be fixed on master?

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,192 @@
...
+
+        // Technically, subscriptions work only if there's one array in the cart
+        if (sizeof($products) > 1)
+        {
+            throw new Exception('Currently Razorpay does not support more than
+                                one product in the cart if one of products
+                                is a subscription.');
+        }
+
+        return array_values($products)[0];
```
mayankamencherla: @captn3m0 Please dump the $products variable. It returns 

```
array (size=1)
  479 => 
    array (size=13)
      'name' => string 'Woo Single #2' (length=13)
      'type' => string 'line_item' (length=9)
      'item_meta' => 
        array (size=9)
          '_qty' => 
            array (size=1)
              ...
          '_tax_class' => 
            array (size=1)
              ...
          '_product_id' => 
            array (size=1)
              ...
          '_variation_id' => 
            array (size=1)
              ...
          '_line_subtotal' => 
            array (size=1)
              ...
          '_line_total' => 
            array (size=1)
              ...
          '_line_subtotal_tax' => 
            array (size=1)
              ...
          '_line_tax' => 
            array (size=1)
              ...
          '_line_tax_data' => 
            array (size=1)
              ...
      'item_meta_array' => 
        array (size=9)
          5335 => 
            object(stdClass)[557]
              ...
          5336 => 
            object(stdClass)[556]
              ...
          5337 => 
            object(stdClass)[555]
              ...
          5338 => 
            object(stdClass)[554]
              ...
          5339 => 
            object(stdClass)[553]
              ...
          5340 => 
            object(stdClass)[552]
              ...
          5341 => 
            object(stdClass)[551]
              ...
          5342 => 
            object(stdClass)[550]
              ...
          5343 => 
            object(stdClass)[549]
              ...
      'qty' => string '1' (length=1)
      'tax_class' => string '' (length=0)
      'product_id' => string '99' (length=2)
      'variation_id' => string '0' (length=1)
      'line_subtotal' => string '8' (length=1)
      'line_total' => string '8' (length=1)
      'line_subtotal_tax' => string '0' (length=1)
      'line_tax' => string '0' (length=1)
      'line_tax_data' => string 'a:2:{s:5:"total";a:0:{}s:8:"subtotal";a:0:{}}' (length=45)
```

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,192 @@
...
+    {
+        $order = new WC_Order($orderId);
+
+        $product = $this->getProduct($order);
+
+        $planId = $this->getProductPlanId($product);
+
+        $customerId = $this->getCustomerId($order);
+
+        $length = (int) WC_Subscriptions_Product::get_length($product['product_id']);
```
mayankamencherla: @captn3m0 As discussed, let us use intval

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,192 @@
...
+
+        $signUpFee = WC_Subscriptions_Product::get_sign_up_fee($product['product_id']);
+
+        if ($signUpFee)
+        {
+            $subscriptionData['addons'] = array(
+                array (
+                    'item' => array(
+                        'amount'   => (int) round($signUpFee * 100, 2),
+                        'currency' => get_woocommerce_currency(),
```
mayankamencherla: @captn3m0 Will do while rebasing

includes/razorpay-webhook.php
```diff
@@ -54,28 +54,45 @@ public function process()
                     write_log(['message' => $e->getMessage()]);
                     return;
                 }
-            }
 
-            switch ($data['event'])
-            {
-                case 'payment.authorized':
-                    return $this->paymentAuthorized($data);
+                switch ($data['event'])
```
mayankamencherla: @captn3m0 Already moved

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Not a fan. That would mean we would be adding an extra parameter in the method call.

razorpay-sdk/src/Utility.php
mayankamencherla: @captn3m0 It is, but we haven't rebased yet.

razorpay-payments.php
mayankamencherla: @captn3m0 Can we discuss this once?

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,246 @@
...
+     */
+    protected function shouldCreatePlan($metadata, $productId)
+    {
+        if (empty($metadata['razorpay_wc_plan_id']) === false)
+        {
+            $planId = $metadata['razorpay_wc_plan_id'][0];
+
+            $plan = $this->api->plan->fetch($planId);
+
+            $salePrice = WC_Subscriptions_product::get_meta_data($productId, 'sale_price', 0) * 100;
```
captn3m0: 100 assumes INR. Needs conversion.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        return $planId;
+    }
+
+    /**
+     * We shouldn't create plan only if the following conditions are met
+     * 1. There exists a plan id as product metadata
+     * 2. The corresponding plan has amount a1
+     * 3. The sale price is amount a2
```
captn3m0: What if the get_price has changed to a3, wouldn't that require us to create a new plan as well?

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,246 @@
...
+
+    protected function createPlan($product)
+    {
+        $productId = $product['product_id'];
+
+        $planArgs = $this->getPlanArguments($product);
+
+        $plan = $this->api->plan->create($planArgs);
+
+        // Storing the plan id as product metadata, unique set to true
```
captn3m0: can you clarify what is meant by unique here?

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,246 @@
...
+        //
+        if (empty($salePrice) === false)
+        {
+            $recurringFee = $salePrice;
+        }
+
+        $planArgs = array(
+            'item' => array(
+                'name'     => $product['name'],
+                'amount'   => (int) round($recurringFee * 100),
```
captn3m0: assumes INR. needs conversion

includes/razorpay-webhook.php
captn3m0: I meant, we missed this in the last release. Which means the present release has a security issue.

includes/razorpay-webhook.php
```diff
@@ -116,4 +135,116 @@ protected function paymentAuthorized($data)
...
+        // This method is used to process the subscription's recurring payment
+        $wcSubscription = wcs_get_subscriptions_for_order($orderId);
+
+        $subscriptionId = get_post_meta($orderId, self::RAZORPAY_SUBSCRIPTION_ID)[0];
+
+        $api = $this->razorpay->getRazorpayApiInstance();
+
+        $subscription = $api->subscription->fetch($subscriptionId);
+
+        $this->processSubscriptionSuccess($wcSubscription, $subscription, $paymentId);
```
captn3m0: How about:

`$this->processSubscriptionSuccess($wcSubscription[0], $subscription[0], $paymentId);`

The method signature asks for `"$subscription"`, which doesn't feel like an array object at all. Much cleaner to pass a required single object rather than an array of exactly one object.

includes/razorpay-webhook.php
```diff
@@ -116,4 +135,116 @@ protected function paymentAuthorized($data)
...
+
+    /**
+     * In the case of successful payment, we mark the subscription successful
+     *
+     * @param $wcSubscription
+     * @param $subscription
+     */
+    protected function processSubscriptionSuccess($wcSubscription, $subscription, $paymentId)
+    {
+        $wcSubscriptionId = array_keys($wcSubscription)[0];
```
captn3m0: Wondering if we need checks around the arrays here. Throw proper errors if count>1 anyhow

captn3m0: Reviewed till here for now. Publishing review. Will review rest asap.

razorpay-sdk/src/Utility.php
```diff
@@ -9,10 +9,20 @@ class Utility
...
         $paymentId = $attributes['razorpay_payment_id'];
 
-        $payload = $orderId . '|' . $paymentId;
+        if (isset($attributes['razorpay_order_id']) === true)
+        {
+            $orderId = $attributes['razorpay_order_id'];
+
+            $payload = $orderId . '|' . $paymentId;
+        }
+        else
```
ccrims0n: use else-if

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Right.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 We aren't processing anything on the `$subscription` object here. We are processing the `$wcSubscription` object. Specifically, we would need the id and the object of the the array. 

Which is why we do the following:

```
$wcSubscriptionId = array_keys($wcSubscription)[0];

// We will only process one subscription per order
$wcSubscription = array_values($wcSubscription)[0];
```

Now, we could process the array in the `processSubscription` method, but personally, I think adding another input parameter makes the line too long. 

Please add your input on this.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 But the return value of when we make a call to `wcs_get_subscriptions_for_order` is of this format. In our case, we're dealing with an array of size 1.

razorpay-sdk/src/Utility.php
mayankamencherla: @ccrims0n I considered that, but, if `razorpay_order_id` isn't set, then `razorpay_subscription_id` will surely be set, and I know that because I wrote the code for it. 

So IMO, it would be better the way it is now.

includes/razorpay-subscriptions.php
mayankamencherla: @captn3m0 add_post_meta(....., true) on line 127. Ensures that the key is unique. 

https://codex.wordpress.org/Function_Reference/add_post_meta

This ensures that the plan id has to be deleted before we add it again.

includes/razorpay-subscriptions.php
mayankamencherla: @captn3m0 Of course. Should add that. But the way our code is structured, I will just add a check for getPrice after salePrice. Logically, that should handle all the edge cases.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,349 @@
...
+        $data = $this->getCustomerInfo($order);
+        $data['fail_existing'] = '0';
+
+        try
+        {
+            $customer = $this->api->customer->create($data);
+        }
+        catch (Exception $e)
+        {
+            echo "Razorpay Error: " . $e->getMessage();
```
captn3m0: try catch doesn't help because you get an echo followed by a fatal error

includes/razorpay-webhook.php
```diff
@@ -91,7 +110,14 @@ protected function paymentAuthorized($data)
...
         $razorpayPaymentId = $data['payload']['payment']['entity']['id'];
 
-        $payment = $this->api->payment->fetch($razorpayPaymentId);
+        try
+        {
+            $payment = $this->api->payment->fetch($razorpayPaymentId);
+        }
+        catch (Exception $e)
+        {
+            echo "Razorpay Error: " . $e->getMessage();
```
captn3m0: echoing in a webhook achieves nothing. Either throw exceptions and catch it somewhere in/above the process function, or log+exit on errors

includes/razorpay-webhook.php
captn3m0: What would be the extra parameter be? Passing single element arrays is not clean, because your function is not easily reusable. If you want to call it elsewhere, for eg, you will be forced to wrap subscription objects as a [$subscription] in the call.

includes/razorpay-webhook.php
captn3m0: Trying to argue for defensive programming. If we ever reach an inconsistent state here (with 2 items in the order, which should never be the case), we should not be processing just the first item

includes/razorpay-webhook.php
```diff
@@ -116,4 +142,123 @@ protected function paymentAuthorized($data)
...
+    {
+        $wcSubscriptionId = array_keys($wcSubscription)[0];
+
+        // We will only process one subscription per order
+        $wcSubscription = array_values($wcSubscription)[0];
+
+        $paymentCount = $wcSubscription->get_completed_payment_count();
+
+        // The subscription is completely paid for
+        if ($paymentCount === $subscription->total_count)
```
captn3m0: @StSarc Need your confirmation on these checks.

razorpay-payments.php
```diff
@@ -292,24 +318,42 @@ protected function getCheckoutArguments($order, $razorpayOrderId)
...
+              'callback_url' => $callbackUrl,
+              'prefill' => $this->getPreFillArguments($order)
             );
 
+            // order_id and subscription_id both cannot be set at the same time
+            if (empty($razorpayOrderId) === false)
+            {
+                $args['order_id'] = $razorpayOrderId;
 
+                $args['amount'] = $this->getOrderAmountAsInteger($order);
```
captn3m0: Discussed with @pronav, we can drop the amount

razorpay-payments.php
```diff
@@ -357,7 +401,14 @@ protected function createRazorpayOrderId($orderId, $sessionKey)
...
             $data = $this->getOrderCreationData($orderId);
 
-            $razorpayOrder = $api->order->create($data);
+            try
+            {
+                $razorpayOrder = $api->order->create($data);
+            }
+            catch (Exception $e)
+            {
+                echo "Razorpay Error: " . $e->getMessage();
```
captn3m0: execution will continue to 413 and raise another exception there, since $razorpayOrder is null.

razorpay-payments.php
```diff
@@ -412,7 +463,14 @@ protected function verifyOrderAmount($razorpayOrderId, $orderId)
...
             $api = $this->getRazorpayApiInstance();
 
-            $razorpayOrder = $api->order->fetch($razorpayOrderId);
+            try
+            {
+                $razorpayOrder = $api->order->fetch($razorpayOrderId);
+            }
+            catch (Exception $e)
+            {
+                echo "Razorpay Error: " . $e->getMessage();
```
captn3m0: similar situation here

razorpay-sdk/src/Utility.php
captn3m0: You can do `if()elseif()else(throw error)`. Makes for more robust code. Not deal breaking, but I'd recommend. This also makes a lot of sense for when we'll be porting these changes to the razorpay-php sdk (where you don't know what the client can/will send across)

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,349 @@
...
+                echo "Razorpay Error: " . $e->getMessage();
+            }
+
+            //
+            // If the product sale price is different from the plan amount
+            // Or if the product price is different from the plan amount
+            // We delete the old plan id from post metadata.
+            // Then, we create a new plan and save id as metadata
+            //
+            if (($this->ifProductOnSale($plan, $productId) === true) or
```
mayankamencherla: @captn3m0 Yes, but there are 2 cases to consider here: 
1. Sale price changed. 
2. Product price changed. 

Both are independent of one another.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,349 @@
...
+        if ((empty($salePrice) === false) and
+            ($salePrice === $plan['item']['amount']))
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    protected function ifProductPriceChanged($plan, $product)
```
captn3m0: nit: hasProductPriceChanged

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,349 @@
...
+            // If we are going to create a new plan id,
+            // we're going to have to delete the old one
+            //
+            delete_post_meta($productId, self::RAZORPAY_PLAN_ID);
+        }
+
+        return true;
+    }
+
+    protected function ifProductOnSale($plan, $productId)
```
captn3m0: nit: isProductOnSale

includes/razorpay-subscriptions.php
captn3m0: I don't understand. Why won't this work:

1. Try to create a new plan and get the arguments as per the current state. This will obviously include whether or not there is a sale and what is the current price
2. Match it against whatever plan we already have on file.

If they don't match, create a new plan. Else use the old one.

mayankamencherla: @captn3m0 You're right. Changed.

includes/razorpay-subscriptions.php
mayankamencherla: @captn3m0 This method is deleted

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 
We would need both 
```
$wcSubscriptionId = array_keys($wcSubscription)[0];

// We will only process one subscription per order
$wcSubscription = array_values($wcSubscription)[0];
```

both to passed here for the method to work. Also, regarding reusability, I agree, but we can say with surety that we process only one subscription at a time. So I think based on that constraint, this would be a valid assumption. 

Let us discuss this once in person and close it.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Logging and exiting.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,349 @@
...
+
+        $subscriptionData = $this->getSubscriptionCreateData($orderId);
+
+        try
+        {
+            $subscription = $this->api->subscription->create($subscriptionData);
+        }
+        catch (Exception $e)
+        {
+            echo "Razorpay Error: " . $e->getMessage();
```
ccrims0n: rethrow exception?

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,349 @@
...
+
+    public function cancelSubscription($subscriptionId)
+    {
+        try
+        {
+            $subscription = $this->api->subscription->cancel($subscriptionId);
+        }
+        catch (Exception $e)
+        {
+            echo "Razorpay Error: " . $e->getMessage();
```
ccrims0n: same here.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    public function getProduct($order)
+    {
+        $products = $order->get_items();
+
+        // Technically, subscriptions work only if there's one array in the cart
+        if (sizeof($products) > 1)
+        {
+            throw new Exception('Currently Razorpay does not support more than
```
ccrims0n: Use `.` for  multi line.

includes/razorpay-subscriptions.php
ccrims0n: @mayankamencherla can you please add the points from discussion here.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,297 @@
...
+
+        return $args;
+    }
+
+    public function getProduct($order)
+    {
+        $products = $order->get_items();
+
+        // Technically, subscriptions work only if there's one array in the cart
+        if (sizeof($products) > 1)
```
ccrims0n: why not use count()?

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,297 @@
...
+                                one product in the cart if one of the products
+                                is a subscription.');
+        }
+
+        return array_values($products)[0];
+    }
+
+    protected function getSubscriptionSessionKey($orderId)
+    {
+        return "razorpay_subscription_id" . $orderId;
```
ccrims0n: use single quotes.

includes/razorpay-webhook.php
ccrims0n: @mayankamencherla why haven't we released a fix for this in separate PR?

includes/razorpay-webhook.php
```diff
@@ -91,7 +110,22 @@ protected function paymentAuthorized($data)
...
+        {
+            $log = array(
+                'message'   => $e->getMessage(),
+                'data'      => $razorpayPaymentId,
+                'event'     => $data['event']
+            );
+
+            write_log($log);
+
+            exit;
```
ccrims0n: @mayankamencherla do a return instead of exit? /cc @captn3m0

includes/razorpay-webhook.php
ccrims0n: I agree with @captn3m0.

razorpay-payments.php
```diff
@@ -257,20 +269,34 @@ protected function createOrGetRazorpayOrderId($orderId)
...
+
+            $redirectUrl = get_site_url() . '/?wc-api=' . get_class($this);
 
-            if(is_a($razorpayOrderId, 'Exception'))
+            if (($razorpayOrderId === null) and
+                (wcs_order_contains_subscription($orderId) === false))
+            {
+                return 'RAZORPAY ERROR: Api could not be reached';
+            }
+            else if(is_a($razorpayOrderId, 'Exception'))
```
ccrims0n: use `instance of`?

razorpay-sdk/src/Utility.php
```diff
@@ -9,10 +9,25 @@ class Utility
...
+        }
+        else if (isset($attributes['razorpay_subscription_id']) === true)
+        {
+            $subscriptionId = $attributes['razorpay_subscription_id'];
+
+            $payload = $paymentId . '|' . $subscriptionId ;
+        }
+        else
+        {
+            throw new Error('Invalid parameters passed to verifyPaymentSignature:
```
ccrims0n: Use `.`.

razorpay-payments.php
```diff
@@ -457,28 +500,32 @@ function getOrderCreationData($orderId)
...
+                );
+            }
+        }
+
         private function enqueueCheckoutScripts($data)
         {
             wp_register_script('razorpay_checkout',
-                'https://checkout.razorpay.com/v1/checkout.js',
-                null, null);
+                'https://checkout.razorpay.com/v1/checkout.js', null, null);
```
ccrims0n: why change this?

razorpay-payments.php
```diff
@@ -457,28 +500,32 @@ function getOrderCreationData($orderId)
...
-                }
+                $this->handleCurrencyConversion($data);
             }
 
             return $data;
         }
 
+        public function handleCurrencyConversion(& $data)
+        {
+            if (class_exists('WOOCS'))
```
ccrims0n: inverse the condition?

razorpay-payments.php
ccrims0n: +1

razorpay-payments.php
ccrims0n: +1

includes/razorpay-subscriptions.php
mayankamencherla: @ccrims0n Returning error message instead. cc @captn3m0

includes/razorpay-subscriptions.php
mayankamencherla: @ccrims0n Returning error message instead. cc @captn3m0

includes/razorpay-subscriptions.php
mayankamencherla: @ccrims0n Does it make a difference? Both achieve the same result, right? Can we leave as is?

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Please add your input here. 

@ccrims0n Typically, we have always stuck to exit instead of return in this plugin.

razorpay-payments.php
mayankamencherla: @ccrims0n Changed.

razorpay-payments.php
mayankamencherla: @ccrims0n 
```
public function handleCurrencyConversion(& $data)
        {
            if (class_exists('WOOCS') === false)
            {
                throw new Errors\BadRequestError(
                    WooErrors\ErrorCode::WOOCS_MISSING_ERROR_MESSAGE,
                    WooErrors\ErrorCode::WOOCS_MISSING_ERROR_CODE,
                    400
                );
            }

            $this->convertCurrency($data);
        }
```

includes/razorpay-webhook.php
captn3m0: discussed with @mayankamencherla. Let us move the 

`$wcSubscription = wcs_get_subscriptions_for_order($orderId);

call to the `processSubscriptionSuccess` function instead

includes/razorpay-webhook.php
captn3m0: @StSarc reminder

includes/razorpay-webhook.php
```diff
@@ -116,4 +150,134 @@ protected function paymentAuthorized($data)
...
+        $wcSubscription = array_values($wcSubscription)[0];
+
+        $paymentCount = $wcSubscription->get_completed_payment_count();
+
+        // The subscription is completely paid for
+        if ($paymentCount === $subscription->total_count)
+        {
+            return;
+        }
+        else if ($wcSubscription->get_completed_payment_count() + 1 === $subscription->paid_count)
```
StSarc: This check shouldn't be required..

includes/razorpay-webhook.php
StSarc: I'll add a `subscription.completed` webhook. Then this check shouldn't be required.

includes/razorpay-webhook.php
mayankamencherla: @StSarc I prefer having this check on here. Although, it may not be required. @captn3m0 could you add your point of view here?

includes/razorpay-webhook.php
mayankamencherla: @StSarc Please let me know when it is done?

includes/razorpay-webhook.php
mayankamencherla: @ccrims0n We should have, but for now released a fix with webhooks disabled by default. This PR will likely get merged today / tomorrow.

includes/razorpay-subscriptions.php
mayankamencherla: @ccrims0n We discussed to leave it as int, and either way does not make a difference.

includes/razorpay-webhook.php
StSarc: This is done.

includes/razorpay-webhook.php
ccrims0n: @mayankamencherla Please do not push multiple things in a single PR.
The fix should have gone live independent of the feature.

includes/razorpay-webhook.php
ccrims0n: @mayankamencherla Can you please add your argument for using `exit`?

includes/razorpay-subscriptions.php
ccrims0n: @mayankamencherla Not really but it's widely known and easier to understand.

includes/razorpay-subscriptions.php
ccrims0n: @mayankamencherla Can you please add the reason for not rethrowing the exception?

razorpay-sdk/src/Utility.php
ccrims0n: @mayankamencherla Why are we making the changing in SDK without actually changing the SDK. This will bring inconsistencies.

includes/razorpay-webhook.php
mayankamencherla: @ccrims0n It should have, but @captn3m0 and I decided that we can push it out together and disabling the webhooks feature.

includes/razorpay-webhook.php
mayankamencherla: @ccrims0n The purpose of exit is to kill the script.

StSarc: Pros of using `exit` over `return`? Is it only convention or is there any specific reason?

mayankamencherla: @StSarc No specific reason as such springs to mind. We were just following that order of things. Is there any specific reason why we should move to return?

includes/razorpay-subscriptions.php
mayankamencherla: @ccrims0n I don't get how it's easier to understand. Small issue, can we leave as is?

includes/razorpay-subscriptions.php
mayankamencherla: @ccrims0n We are doing this everywhere in the plugin

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Remove this check?

razorpay-sdk/src/Utility.php
mayankamencherla: @ccrims0n We'll change the SDK too. But this will likely go live before the SDK change.

includes/razorpay-webhook.php
captn3m0: @StSarc if we do an exit, other hooks can't run. Since it was a razorpay webhook and couldn't be authenticated we don't want any further processing in this request.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,299 @@
...
+    {
+        $products = $order->get_items();
+
+        //
+        // Technically, subscriptions work only if there's one array in the cart
+        //
+        if (sizeof($products) > 1)
+        {
+            throw new Exception('Currently Razorpay does not support more than'
+                                . 'one product in the cart if one of the products'
```
ccrims0n: space after `'`

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,372 @@
...
+
+        $currency = get_woocommerce_currency();
+
+        $key = self::RAZORPAY_PLAN_ID . '_'. strtolower($currency);
+
+        //
+        // If razorpay_plan_id is set in the metadata,
+        // we check if the amounts match and return the plan id
+        //
+        if (isset($metadata[$key]) === true)
```
captn3m0: refactor this so that there is a single createPlan call in this method.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,372 @@
...
+        return $plan['id'];
+    }
+
+    protected function getPlanArguments($product)
+    {
+        $productId = $product['product_id'];
+
+        $period       = WC_Subscriptions_Product::get_period($productId);
+        $interval     = WC_Subscriptions_Product::get_interval($productId);
+        $salePrice    = get_post_meta($productId)['_sale_price'][0];
```
captn3m0: no better way to get sale price from a product?

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,372 @@
...
+    {
+        $currency = get_woocommerce_currency();
+
+        $key = self::RAZORPAY_PLAN_ID . '_'. strtolower($currency);
+
+        $productId = $product['product_id'];
+
+        $metadata = get_post_meta($productId);
+
+        list($planId, $created) = $this->createOrGetPlanId($metadata, $product);
```
captn3m0: suggestion: pass currency to this method, so that we don't have to fetch it again. Makes it more deterministic.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,372 @@
...
+        $interval     = WC_Subscriptions_Product::get_interval($productId);
+        $salePrice    = get_post_meta($productId)['_sale_price'][0];
+        $recurringFee = WC_Subscriptions_Product::get_price($productId);
+
+        //
+        // If the item is on sale, sell it for the sale price
+        //
+        if (empty($salePrice) === false)
+        {
+            $recurringFee = $salePrice;
```
captn3m0: Use something like https://stackoverflow.com/a/43966170/368328 instead.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,372 @@
...
+            'year'  => 'yearly'
+        );
+
+        return $periodMap[$period];
+    }
+
+    protected function getCustomerId($order)
+    {
+        $data = $this->getCustomerInfo($order);
+        $data['fail_existing'] = '0';
```
captn3m0: Give a link to the [docs](https://docs.razorpay.com/v1/page/customers-api) so it is clear what this does.

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,372 @@
...
+                $message,
+                WooErrors\ErrorCode::API_CUSTOMER_CREATION_FAILED,
+                400
+            );
+        }
+
+        return $customer['id'];
+    }
+
+    protected function getCustomerInfo($order)
```
captn3m0: rename `getPreFillArguments` to `getCustomerDetails` and mark it as public static and reuse here?

includes/razorpay-subscriptions.php
```diff
@@ -0,0 +1,372 @@
...
+                'name'    => $order->billing_first_name . ' ' . $order->billing_last_name,
+                'email'   => $order->billing_email,
+                'contact' => $order->billing_phone,
+            );
+        }
+
+        return $args;
+    }
+
+    public function getProduct($order)
```
captn3m0: getProductFromOrder

includes/razorpay-webhook.php
```diff
@@ -110,10 +144,162 @@ protected function paymentAuthorized($data)
...
+     */
+    protected function processSubscriptionFailed($orderId)
+    {
+        WC_Subscriptions_Manager::process_subscription_payment_failure_on_order($orderId);
+    }
+
+    /**
+     * Returns the order amount, rounded as integer
+     */
+    public function getOrderAmountAsInteger($order)
```
captn3m0: Why redefine?

includes/razorpay-webhook.php
```diff
@@ -110,10 +144,162 @@ protected function paymentAuthorized($data)
...
+     * Currently we handle only subscription failures using this webhook
+     *
+     * @param $data
+     */
+    protected function paymentFailed($data)
+    {
+        //
+        // Order entity should be sent as part of the webhook payload
+        //
+        $orderId = $data['payload']['payment']['entity']['notes']['woocommerce_order_id'];
```
captn3m0: note for a future release: this will break if merchant is using the same account on another website without orders API

includes/razorpay-subscriptions.php
mayankamencherla: @captn3m0 How about setting $this->currency during the constructor?

includes/razorpay-subscriptions.php
mayankamencherla: @captn3m0 This was the only one that worked on the ken. Let me re-check.

includes/razorpay-subscriptions.php
mayankamencherla: @captn3m0 `is_on_sale` etc not available on subscriptions products

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Haven't defined it anywhere else. Could you be more specific?

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 But then won't the url be different? All woocommerce url's will be used with orders api, because we know our plugin uses orders api.

includes/razorpay-subscriptions.php
captn3m0: It is confusing because you are generating the key twice. Pass the key instead?

includes/razorpay-subscriptions.php
captn3m0: @mayankamencherla `get_price` takes care of this for you: https://github.com/wp-premium/woocommerce-subscriptions/blob/master/includes/class-wc-subscriptions-product.php#L384-L404

includes/razorpay-webhook.php
captn3m0: Single webhook is used for the entire merchant account.

includes/razorpay-webhook.php
captn3m0: Defined on the payments class.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 I get that, but I mean, this code will be triggered only if webhook url is set to the woocommerce url. So I think we're fine here.

includes/razorpay-webhook.php
mayankamencherla: @captn3m0 Not anymore.

includes/razorpay-subscriptions.php
mayankamencherla: @captn3m0 Which key?

PR #46, opened 2017-06-23
Add support for currency switcher plugin

albingeorge:
```
If the currency is not INR:
      If the currency swicth plugin is installed
        If both INR and current currency is configured in the currency
        switch plugin:
          Convert to INR and process further
        Else
          Throw an error asking to configure the CS plugin with both INR and
          the current currency
    Else
      Throw an error saying currency is not supported
```


razorpay-sdk/src/Errors/ErrorCode.php
```diff
@@ -4,9 +4,18 @@
 
 class ErrorCode
 {
-    const BAD_REQUEST_ERROR                 = 'BAD_REQUEST_ERROR';
-    const SERVER_ERROR                      = 'SERVER_ERROR';
-    const GATEWAY_ERROR                     = 'GATEWAY_ERROR';
+    const BAD_REQUEST_ERROR                    = 'BAD_REQUEST_ERROR';
```
captn3m0: Create a new class for this. Try not to make any changes to sdk that we can't push back upstream.

razorpay-sdk/src/Errors/InvalidCurrencyError.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Razorpay\Api\Errors;
+
+class InvalidCurrencyError extends Error
```
captn3m0: Put this in a different directory and namespace?

razorpay-payments.php
```diff
@@ -329,13 +335,94 @@ protected function createRazorpayOrderId($orderId, $sessionKey)
...
+        }
+
+        /**
+         * Validate the order creation data before calling the order create API
+         *
+         * @param  Array $data The data passed to create order API
+         *
+         * @return void
+         */
+        protected function validateOrderCreateData($data): Array
```
captn3m0: The flow here seems to be using exceptions pretty loosely. Instead of:

1. Setup data
2. Validate (if, check, throw)
3. Catch-> Fix
4. Proceed

Why not just check the conditions upfront:

```
if(INR) ->
else if ($installed and not INR)
else if(not INR)
```

razorpay-payments.php
```diff
@@ -329,13 +341,68 @@ protected function createRazorpayOrderId($orderId, $sessionKey)
...
+                $data['amount'] = round(
+                    (($data['amount'] * $currencies['INR']['rate']) / ($currencies[$data['currency']]['rate'] == 0 ? 1 : $currencies[$data['currency']]['rate'])),
+                    0
+                );
+                $data['currency'] = 'INR';
+                return $data;
+            }
+            else
+            {
+                throw new Errors\BadRequestError(
```
captn3m0: How does this show up during a payment attempt?

razorpay-payments.php
```diff
@@ -329,13 +341,68 @@ protected function createRazorpayOrderId($orderId, $sessionKey)
...
+        {
+            global $WOOCS;
+
+            $currencies = $WOOCS->get_currencies();
+
+            $value = $data['amount'] * $currencies[$WOOCS->current_currency]['rate'];
+
+            if (array_key_exists('INR', $currencies) and array_key_exists($data['currency'], $currencies))
+            {
+                // Convert the currency to INR using the rates fetched from the Currency Switcher plugin
```
albingeorge: Added the instructions to get this working on the above wiki page.

albingeorge:
Screenshots

1. Currency set from WooCommerce: GBP
    Currency Switcher plugin: Not installed

plugin_missing


2. Currency set from WooCommerce: USD
    Currency Switcher plugin: Installed, INR not added

currency_not_configured


3. Currency set from WooCommerce: USD
    Currency Switcher plugin: Installed and configured

order_success

razorpay-payments.php
albingeorge: I've added screenshots below.

captn3m0:
Can we improve this wording? Saying pay button where none exists doesn't look good.

image

Same for the first case as well.

albingeorge:
@captn3m0 Can we hide the message 'Thank you for your order, please click the button below to pay with Razorpay.' based on whether there is an error shown?

Or should we change the verbiage of the error message to something simpler like 'RAZORPAY ERROR: Payment failed'.

Previously, we used to show 'RAZORPAY ERROR: Api could not be reached' for every payment failures(even if the API was reachable).

captn3m0:
>Can we hide the message

Yup, sounds like a good idea.

>Previously, we used to show 'RAZORPAY ERROR: Api could not be reached' for every payment failures(even if the API was reachable).

The idea was not to disclose anything sensitive by accident. If we are certain that nothing sensitive will be revealed by the exception message (ie, we are only catching known exceptions, and any other generic exception gets a generic message), I'm happy with switching this to detailed errors.

albingeorge:
Regarding the catching of exception, as of now, we're catching `Exception` and the message from the same is pushed to the front-end.

Instead, we can catch only the `BadRequestError` exception and for the rest of the exceptions, we can show some generic message like 'RAZORPAY ERROR: Payment failed'. Let me know if that works.

captn3m0:
Yup, that is what I was suggesting as well :+1:

razorpay-payments.php
```diff
@@ -275,17 +285,25 @@ protected function getCheckoutArguments($order, $razorpayOrderId)
...
-              'callback_url' => $callbackUrl,
+              'key'              => $this->key_id,
+              'name'             => get_bloginfo('name'),
+              // Harcoding currency to INR, since if not INR, an exception gets thrown
+              'currency'         => 'INR',
+              // 'currency'     => get_woocommerce_currency(),
+              'description'      => $productinfo,
+              'notes'            => array(
+                                        'woocommerce_order_id' => $orderId
+                                    ),
```
captn3m0: closing bracket should align on left

razorpay-payments.php
```diff
@@ -329,13 +347,73 @@ protected function createRazorpayOrderId($orderId, $sessionKey)
...
+            $currencies = $WOOCS->get_currencies();
+
+            $value = $data['amount'] * $currencies[$WOOCS->current_currency]['rate'];
+
+            if (array_key_exists('INR', $currencies) and array_key_exists($data['currency'], $currencies))
+            {
+                // If the currenct currency is the same as the default currency set in WooCommerce,
+                // Currency Switcher plugin sets the rate of currenct currency as 0, because of which
+                // we need to set this to 1 here if it's value is 0
+                $current_currency_rate = ($currencies[$data['currency']]['rate'] == 0 ? 1 : $currencies[$data['currency']]['rate']);
```
captn3m0: camelCase everywhere, unless wordpress hook-related things

PR #48, opened 2017-07-01
Release 1.5.0

captn3m0:
@albingeorge: Made changes regarding order caching. We were creating double orders for every refresh on non-INR orders. There were also rounding errors that I was getting, so fixed those as well.

@mayankamencherla: There was a security issue with how we were display error messages without sanitization.


albingeorge:
LGTM. Also tested locally.

PR #49, opened 2017-07-05
fix: Made backward compatible for getting currency

albingeorge:
Use get_currency() or get_order_currency() based on the WooCommerce version.

Tested against the WooCommerce versions 3.1.0 and 2.6.8.

### DIFF
--- a/razorpay-payments.php
+++ b/razorpay-payments.php
@@ -289,6 +289,8 @@ function woocommerce_razorpay_init()
 
             $productinfo = "Order $orderId";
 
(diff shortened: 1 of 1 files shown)


captn3m0:
:+1: for the quick fix. 

I'd have preferred a version_compare check though, like we do it in other places.

albingeorge:
@captn3m0 I used the same version_compare block we use in prefill arguments for this.

captn3m0:
Oops, was not showing in the default diff. :+1:

PR #62, opened 2017-09-28
Split sub bug

mayankamencherla:
Returning if the payment authorized contains a subscription payment


includes/razorpay-webhook.php
```diff
@@ -106,19 +106,26 @@ protected function paymentFailed(array $data)
      */
     protected function paymentAuthorized(array $data)
     {
+        // We don't process subscription payments here
+        if (isset($data['payload']['payment']['entity']['invoice_id']) === true)
+        {
+            $invoiceId = $data['payload']['payment']['entity']['invoice_id'];
```
captn3m0: Ignore as soon as we notice an invoice id, since we can't process any invoice payments here (with or without subscriptions doesn't matter)

PR #65, opened 2017-09-29
[order id] Using order->get_order_number instead of order->get_id

mayankamencherla:
Using order->get_order_number instead of order->get_id


captn3m0:
This affects the note parameter `woocommerce_order_id`, which would be a breaking change.

PR #70, opened 2017-11-03
Initial travis build

captn3m0:


tessie:
@captn3m0 
This is done so far. What is pending

1 . Test  functions that need to mock API eg refunds
2 . Need to set up currency plugin

captn3m0:
- We require atleast Wordpress 3.9.2, so testing below that doesn't get us anything.
- Try adding coveralls first, so we have an idea of the coverage we are at
- Ignore methods that call the API for now
- Would be nice if we can test the currency converter for all cases, as this is where we've seen bugs in the past.

tessie:
sorry my mistake. I meant wordpress 3.9.2  Not 2.x versions.  wordpress 3.9.2 and woocommerce latest version has some issues. For wordpress 3.9.2  The latest version of woocommerce is not supported. Have to load woocommerce 2.4

captn3m0:
Let us get it working and merged with the latest versions of everything, and we can work on adding support for older releases later?

tessie:
@nemo 
   - Currency conversion is tested 
   -  Refund for blank refunds is tested
   
Issues 
  Read environment variables . on running coveralls

tessie:
coverage - 47.12%

PR #73, opened 2017-11-20
[webhook capture fix] Capturing payment only if auto capture is not true

mayankamencherla:
- Currently we are fetching a payment during the webhook flow
- We check if the payment status is captured
- If it is not, and the user wants the payment auto-captured, we attempt to capture it
- But if it is captured in the meanwhile, we we catch the exception
- We re fetch the payment entity in the catch block and check for captured status
- If it is captured, we mark the payment as successful

### DIFF
--- a/includes/razorpay-webhook.php
+++ b/includes/razorpay-webhook.php
@@ -143,22 +143,7 @@ class RZP_Webhook
 
         $razorpayPaymentId = $data['payload']['payment']['entity']['id'];
 
-        try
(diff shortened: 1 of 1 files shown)


mayankamencherla:
@captn3m0 This has been tested. Used ngrok, made a payment only via real webhook, and manually coded payment['status'] = 'authorized' to ensure exception is thrown and ensured that order status got updated correctly to true.

PR #78, opened 2017-12-07
Allow merchants to customize order successful message

tessie:


razorpay-payments.php
```diff
@@ -48,6 +48,8 @@ class WC_Razorpay extends WC_Payment_Gateway
 
         const DEFAULT_LABEL                  = 'Credit Card/Debit Card/NetBanking';
         const DEFAULT_DESCRIPTION            = 'Pay securely by Credit or Debit card or Internet Banking through Razorpay.';
+        const DEFAULT_SUCCESS_MESSAGE         = 'Thank you for shopping with us. Your account has been charged and your transaction is successful. We will be
```
ccrims0n: @tessie We should do it via i18n (`_()`)

razorpay-payments.php
```diff
@@ -195,6 +198,12 @@ public function init_form_fields()
                         self::CAPTURE   => 'Authorize and Capture'
                     )
                 ),
+                'order_success_message' => array(
+                    'title' => _('Order Completion Message', $this->id),
```
ccrims0n: function name is double underscore.

razorpay-payments.php
```diff
@@ -48,6 +48,8 @@ class WC_Razorpay extends WC_Payment_Gateway
 
         const DEFAULT_LABEL                  = 'Credit Card/Debit Card/NetBanking';
         const DEFAULT_DESCRIPTION            = 'Pay securely by Credit or Debit card or Internet Banking through Razorpay.';
+        const DEFAULT_SUCCESS_MESSAGE  = 'Thank you for shopping with us. Your account has been charged and your
```
tessie: @ccrims0n  We cannot add the i8n method here.instead I have used in line number 205.

PR #93, opened 2018-12-05
Add support for Price Based on Country multicurrency

oscargare:
Hi, 
I added support for Price Based on Country for WooCommerce plugin. The code converts the currency amount to INR when Price Based on Country is active.

Thanks!

### DIFF
--- a/razorpay-payments.php
+++ b/razorpay-payments.php
@@ -622,16 +622,23 @@ function woocommerce_razorpay_init()
 
         public function handleCurrencyConversion(& $data)
         {
-            if (class_exists('WOOCS') === false)
-            {
+			if (class_exists('WOOCS'))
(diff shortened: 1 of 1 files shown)


srameshr:
@shashankkumar Could you please merge it and update the plugin?

captn3m0:
Thanks a lot for the PR. Since we've moved away from Currency Switcher in the new release, this will need some changes before it can be merged.

captn3m0:
I've tested the plugin with the Price Based on Country plugin, and it works great without any modifications.

Since it overrides the Order method for get_currency and get_total, it works correctly on its own. Thanks @oscargare for the PR. I'll be closing this and notify you once the supported version is released (so you can mark it as supported).

PR #113, opened 2019-06-24
PAY-1440: Pass version info in checkout options

ramth05:
Added version info variables in checkout options.

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -17,6 +17,7 @@ if ( ! defined( 'ABSPATH' ) )
 
 require_once __DIR__.'/includes/razorpay-webhook.php';
 require_once __DIR__.'/razorpay-sdk/Razorpay.php';
+require_once(ABSPATH.'wp-admin/includes/plugin.php');
 
 use Razorpay\Api\Api;
 use Razorpay\Api\Errors;
@@ -567,6 +568,14 @@ function woocommerce_razorpay_init()
 
             wp_register_script('razorpay_wc_script', plugin_dir_url(__FILE__)  . 'script.js',
                 array('razorpay_checkout'));
+
+            $mod_version = get_plugin_data(plugin_dir_path(__FILE__).'woo-razorpay.php')['Version'];
+
+            $data['_'] = array(
+                    'integration'           => 'woocommerce',
+                    'integration_version'   => $mod_version,
+                    'integration_parent_version' => WOOCOMMERCE_VERSION,
+            );
 
             wp_localize_script('razorpay_wc_script',
                 'razorpay_wc_checkout_vars',
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -17,7 +17,7 @@ if ( ! defined( 'ABSPATH' ) )
 
 require_once __DIR__.'/includes/razorpay-webhook.php';
 require_once __DIR__.'/razorpay-sdk/Razorpay.php';
-require_once(ABSPATH.'wp-admin/includes/plugin.php');
+require_once ABSPATH . 'wp-admin/includes/plugin.php';
 
 use Razorpay\Api\Api;
 use Razorpay\Api\Errors;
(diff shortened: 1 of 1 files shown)


woo-razorpay.php
```diff
@@ -17,6 +17,7 @@
 
 require_once __DIR__.'/includes/razorpay-webhook.php';
 require_once __DIR__.'/razorpay-sdk/Razorpay.php';
+require_once(ABSPATH.'wp-admin/includes/plugin.php');
```
harman28: Could you remove the parentheses for consistency with the above imports?

woo-razorpay.php
```diff
@@ -568,6 +569,14 @@ private function enqueueCheckoutScripts($data)
             wp_register_script('razorpay_wc_script', plugin_dir_url(__FILE__)  . 'script.js',
                 array('razorpay_checkout'));
 
+            $mod_version = get_plugin_data(plugin_dir_path(__FILE__).'woo-razorpay.php')['Version'];
```
captn3m0: nit: space around periods

woo-razorpay.php
captn3m0: nit: no () for require_once.

woo-razorpay.php
```diff
@@ -568,6 +569,14 @@ private function enqueueCheckoutScripts($data)
...
+
+            $data['_'] = array(
+                    'integration'           => 'woocommerce',
+                    'integration_version'   => $mod_version,
```
harman28: Align the rockets 🚀

woo-razorpay.php
```diff
@@ -568,6 +569,14 @@ private function enqueueCheckoutScripts($data)
             wp_register_script('razorpay_wc_script', plugin_dir_url(__FILE__)  . 'script.js',
                 array('razorpay_checkout'));
 
+            $mod_version = get_plugin_data(plugin_dir_path(__FILE__).'woo-razorpay.php')['Version'];
+
+            $data['_'] = array(
```
harman28: I feel it is more appropriate to move this to `getDefaultCheckoutArguments` and not here in `enqueueCheckoutScripts`. Please let me know if I'm missing something.

woo-razorpay.php
```diff
@@ -568,6 +569,14 @@ private function enqueueCheckoutScripts($data)
...
+                    'integration'           => 'woocommerce',
+                    'integration_version'   => $mod_version,
+                    'integration_parent_version' => WOOCOMMERCE_VERSION,
```
captn3m0: Can you confirm if this is defined in all of our supported versions?

woo-razorpay.php
ramth05: yes, its fine

woo-razorpay.php
ramth05: ok

woo-razorpay.php
ramth05: yep

woo-razorpay.php
ramth05: This "WOOCOMMERCE_VERSION" const is present in first release as well, so it's supported by all of our plugins versions.

woo-razorpay.php
ramth05: ok

woo-razorpay.php
```diff
@@ -491,9 +498,7 @@ protected function createRazorpayOrderId($orderId, $sessionKey)
             }
             catch (Exception $e)
             {
-                $message = $e->getMessage();
-
-                return 'RAZORPAY ERROR: Order creation failed with the message \'' . $message . '\'';
+                return $e;
```
harman28: This will catch the exception and return it. Won't fix the problem of the exception being caught.

ramth05: it's getting caught in parent calling "getRazorpayPaymentParams" function.

harman28: That will also work if you just call `$api->order->create($data)` directly. What is the need for the try-catch and return?

ramth05: It's written to handle the error in proper way and display the message properly.

PR #132, opened 2019-11-08
PON-39 fixes the cart reset issue in hosted checkout

ChetanGN:


woo-razorpay.php
```diff
@@ -770,6 +771,16 @@ function check_razorpay_response()
                 $this->redirectUser($order);
             }
 
+            if(empty($_POST[self::RAZORPAY_PAYMENT_ID]))
```
ramth05: This will change the behaviour of normal checkout flow.

PR #153, opened 2020-06-03
Added check point for refund webhook

prem-techchefs:


includes/razorpay-webhook.php
```diff
@@ -271,8 +271,16 @@ public function refundedCreated(array $data)
             return;
         }
 
+        //Avoid to recreate refund, If already refund save if woocommerce.
```
ramth05, line 274: //Avoid to recreate refund, If already refund saved and initiated from woocommerce website.

prem-techchefs: Just for comment section, should I remove this?

PR #165, opened 2020-08-24
Order Number != Order ID

prathamesh-gharat:
Order Number is equal to Order ID by default but that might not always be the case.

Following code expects an Order ID and not an Order Number.
https://github.com/KDSPL/razorpay-woocommerce/blob/master/includes/razorpay-webhook.php#L158

Ref: https://docs.woocommerce.com/document/managing-orders/
> Order IDs are non-sequential as they use the default WordPress ID approach. For sequential order numbers, you can use Sequential Order Numbers Pro.

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -409,9 +409,10 @@ function woocommerce_razorpay_init()
         {
             $callbackUrl = $this->getRedirectUrl();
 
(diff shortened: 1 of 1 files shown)


prathamesh-gharat:
Any update on this?

PR #173, opened 2020-12-29
Use get_option to prevent undefined notice

moon0326:
Hi 👋 

I'm one of WC Admin plugin devs and we're trying to integrate Razorpay into the upcoming WooCommerce 4.9.

During the WooCommerce onboarding process, we ask users to enter only `key_id` and `key_secret` as the other values such as title and description are defined in the Razorpay plugin already.

After completing the onboarding process, I received the following PHP notice.

`Undefined index: title in /Users/moonkuykyong/Dev/woocommerce-dev-docker/wordpress/wp-content/plugins/woo-razorpay/woo-razorpay.php on line 119`

There is an [open issue](https://github.com/woocommerce/woocommerce-admin/issues/5974) related to this in WC github.

This PR fixes the issue by using [get_option](https://github.com/woocommerce/woocommerce/blob/b3be8931d641e6dddd349280493b415726de2bec/includes/abstracts/abstract-wc-settings-api.php#L280) function, which respects the default values defined in the [init_form_fields](https://github.com/razorpay/razorpay-woocommerce/blob/22def9f9bc91b78e9a9944b6bdf7443b0be02a5d/woo-razorpay.php#L174) function.

Thank you,
Moon

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -116,7 +116,7 @@ function woocommerce_razorpay_init()
          */
         public function getSetting($key)
         {
-            return $this->settings[$key];
+            return $this->get_option($key);
         }
 
         protected function getCustomOrdercreationMessage()


ramth05 (approved):
LGTM

moon0326:
Thank you @ramth05 👍 Do you have ETA for the release?

Issue #176, opened 2021-02-20
Store the Razor pay order_id against the WooCommerce order_id?


prathamesh-gharat:
Razor pay APIs only have a way to fetch order details using the `order_xxxxxxxxxxxx` IDs.

Is there a specific reason for not saving the Razor pay order_id against the WooCommerce order_id (in post meta)?

PR #178, opened 2021-02-22
Store Razorpay order_id in WooCommerce order meta

prathamesh-gharat:
Since the Razor pay API only allows users to reference orders using the Razor pay order_id it would be useful to retain the Razorpay order_id in WooCommerce for future reference. For example, in an order related automation / order sync process.

#176


prathamesh-gharat:
Accidentally added an older commit

PR #185, opened 2021-05-05
Used transient method to store temp data.

ChetanGN:


ramth05 (commented):
@ChetanGN , Can we use DB to store/fetch rzp orderId. And changing the method is not the good way to fix. why session is lost after 1 or 2 attempt we need to find root cause for it.

ChetanGN:
@ramth05 I feel it's not necessary to store these details in DB for temporary use. For these kinds of temporary data storage, WP is giving a **transient method** which is useful in this case. And about session loss not able to find the exact cause for it.

jampulanaveen (approved):
@ramth05 @ChetanGN i have tested with **transient method** its working fine, and also tested with session but not able to find exact reason why its getting losted.

ramth05:
> @ramth05 @ChetanGN i have tested with **transient method** its working fine, and also tested with session but not able to find exact reason why its getting losted.

@ChetanGN is this **transient method* method save data on server? then what is it's life span?

ChetanGN:
@ramth05 By default there is no expiration. We can set the expiration time in seconds. 
https://developer.wordpress.org/reference/functions/set_transient/

ChetanGN:
@jampulanaveen Can you please test this PR in your local setup.

jampulanaveen (commented):
LGTM

PR #186, opened 2021-05-09
Updated the wc order fetch syntax to wc_get_order()

ChetanGN:


ramth05 (approved):
LGTM. @ChetanGN let it be tested properly. @HancilSequeira , Can you please help us to do quick test.

HancilSequeira (approved):

PR #193, opened 2021-05-19
Sending order number in notes instead of order id.

ChetanGN:
This will help to map the woocommerce order number with the razorpay dashboard even if the custom order number is used.


ramth05 (changes requested):
Please verify the requested change

ramth05 (approved):
LGTM, @ChetanGN , Please release with proper details, what changed. So that the any existing integration will not break.

PR #226, opened 2022-02-23
disable coupon on discount rule

poojarazor:


includes/api/coupon-get.php
```diff
@@ -67,6 +67,14 @@ function getCouponList($request)
         'posts_per_page' => -1, // By default WP_Query will return only 10 posts, to avoid that we need to pass -1
     );
 
+    //check woo-discount-rule plugin disabling the coupons
+    $discountOptions = get_option( 'woo-discount-config-v2', array());
```
ChetanGN, line 71: What if the plugin itself not installed? Is this option available on `get_option` function? @poojarazor

includes/api/coupon-get.php
```diff
@@ -67,6 +67,14 @@ function getCouponList($request)
...
+    $isCouponEnabled = $discountOptions['disable_coupon_when_rule_applied'];
```
ChetanGN, line 72: Incase of `get_option( 'woo-discount-config-v2', array());` fails or empty array, i guess it will throw an error here @poojarazor

ChetanGN: Error handling is missing.

poojarazor: Incase of empty array it will not throw any error, it will set as a  null value

includes/api/coupon-get.php
poojarazor: This option will not available when plugin not installed , value set as a null value

varun-315: Consider using `[]` over `array()` based on style guide - Chetan please help

includes/api/coupon-get.php
ChetanGN: @poojarazor Can user apply the coupon manually?

ChetanGN (approved):
LGTM

PR #239, opened 2022-03-17
qty issue for puca theme

poojarazor:


templates/rzp-pdp-checkout-btn.php
```diff
@@ -22,15 +22,23 @@ class="button alt single_add_to_cart_button"
 </div>
 
 <script type="">
-  var quantity = document.getElementsByClassName("qty")[0].value;
+  let i = 0;
+  while (typeof quantity === 'undefined') {
```
varun-315, line 26: why is a `while` loop needed? Won't buy now only have single item and quantity field?

ChetanGN: @varun-315 Some themes will not have the qty info on the 0th index so we need to loop it through to get the right qty selected.

ChetanGN (approved):
LGTM

ChetanGN:
@poojarazor I hope you had tested these changes with other themes also. It should not break the flow for other themes.

PR #243, opened 2022-03-31
enable coupon field

poojarazor:


woo-razorpay.php
```diff
@@ -1873,26 +1873,8 @@ function addAdminSettingsAlertScript()
 
 add_action('admin_enqueue_scripts', 'addAdminSettingsAlertScript');
 
-function disable_coupon_field_on_cart($enabled)
```
ChetanGN, line 1876: @poojarazor Please let the coupon apply field be hidden on the cart page. You can disable it on the checkout page

poojarazor: got it

ChetanGN (approved):
LGTM

PR #248, opened 2022-04-07
add dual checkout button

poojarazor:


ChetanGN:
@poojarazor Please don't apply beautify on woo-razorpay.php file. Revert it and push again.

ChetanGN:
@poojarazor Also take a pull from master branch before raising PR. Now we had moved 3.2 version.

woo-razorpay.php
```diff
@@ -1778,123 +1608,120 @@ function razorpay_webhook_init()
...
 //To add 1CC button on cart page.
-add_action( 'woocommerce_proceed_to_checkout', 'addCheckoutButton');
+add_action('woocommerce_proceed_to_checkout', 'addCheckoutButton');
 
 function addCheckoutButton()
 {
-  add_action('wp_enqueue_scripts', 'enqueueScriptsFor1cc', 0);
+    add_action('wp_enqueue_scripts', 'enqueueScriptsFor1cc', 0);
 
-  if (isRazorpayPluginEnabled() && is1ccEnabled() )
```
ChetanGN, line 1821: @poojarazor why you had removed this condition? Merchant has to enable the 1cc & plugin has to be activated right!! if you remove this condition then even if merchant has disabled the 1cc from wc dashboard, it will show magic checkout on cart page right!

woo-razorpay.php
```diff
@@ -1778,123 +1608,120 @@ function razorpay_webhook_init()
...
-  {
-    if (isTestModeEnabled()) {
-      $current_user = wp_get_current_user();
-      if ($current_user->has_cap( 'administrator' ) || preg_match( '/@razorpay.com$/i', $current_user->user_email )) {
+    //TODO add flag for dual magic chekout option
+    if (1 == 1) {
```
ChetanGN, line 1650: Make condition proper. Define one variable where $dual_checkout = true; and use that variable here instead of (1==1) @poojarazor

woo-razorpay.php
poojarazor: Haven’t removed from here because of version mismatch removed from here and added below

PR #256, opened 2022-04-20
1cc order draft

ChetanGN:


includes/api/order.php
```diff
@@ -95,6 +95,12 @@ function createWcOrder(WP_REST_Request $request)
         if (is_wp_error($orderId)) {
             $checkout_error = $orderId->get_error_message();
         }
+        //Keep order in draft status untill customer info available
+        wp_update_post(array(
```
varun-315, line 99: Can we add this status in the initial create order? Any new db update may add lot of delay for slow merchants

woo-razorpay.php
```diff
@@ -1188,7 +1193,7 @@ function check_razorpay_response()
                     $razorpayOrderId = get_transient($sessionKey);
                     $razorpayData = $api->order->fetch($razorpayOrderId);
 
-                    $this->updateOrderAddress($razorpayData['items'][0], $order);
+                    $this->updateOrderAddress($razorpayData, $order);
```
varun-315, line 1196: What is this? Looks like it is coming from a contract change?

woo-razorpay.php
```diff
@@ -1102,6 +1102,11 @@ function check_razorpay_response()
             if (count($postIds) > 0)
             {
                 $orderId = $postIds[0];
+
+                wp_update_post(array(
```
varun-315, line 1106: Consider a new function `updateOrderStatus(string orderId, string status) { return ... }

poojarazor (approved):
LGTM

includes/api/order.php
ChetanGN: During the order creation by default, wooc will set the status to wc-pending. No option is available to create it in draft status.

PR #261, opened 2022-04-29
added condition to check object exists

jampulanaveen:


includes/razorpay-webhook.php
```diff
@@ -485,4 +506,15 @@ public function refundedCreated(array $data)
         // Graceful exit since payment is now refunded.
         exit();
     }
+
+    public function checkIsObject($order)
```
ramth05, line 510: @jampulanaveen , please move `wc_get_order` under this only.

includes/razorpay-webhook.php
```diff
@@ -292,6 +299,13 @@ protected function virtualAccountCredited(array $data)
 
         $order = wc_get_order($orderId);
 
+        if(!empty($orderId))
```
ramth05, line 302: @jampulanaveen , this need be validated before calling `wc_get_order`

nikhilsharma1100 (approved):

PR #274, opened 2022-05-18
buf fix

jampulanaveen:


woo-razorpay.php
```diff
@@ -588,13 +588,18 @@ private function getRedirectUrl($orderId)
         protected function getRazorpayPaymentParams($orderId)
         {
             $getWebhookFlag =  get_option('webhook_enable_flag');
-
+            $time = time();
            if (!empty($getWebhookFlag))
            {
                 if ($getWebhookFlag + 86400 < time())
```
ChetanGN, line 594: @ramth05 Is this condition correct? Are we verifying within 24 ours or after 24 hours? cc// @jampulanaveen

ramth05: @ChetanGN , yes it's correct. If we don't have time value when last webhook enabled/updated. We need to update it with the transaction. 
if it's already set then verify the 24 hours timeframe and verify the webhook again.

woo-razorpay.php
```diff
@@ -588,13 +588,18 @@ private function getRedirectUrl($orderId)
...
                 {
                     $this->autoEnableWebhook(); 
                 }
+           }
+           else
+           {
+                add_option('webhook_enable_flag', $time);
```
ChetanGN, line 601: @jampulanaveen `add_option` is already called in `autoEnableWebhook` function, why again it's required here?
`if (empty($getWebhookFlag))
            {
                add_option('webhook_enable_flag', $time);
            }
            else
            {
                update_option('webhook_enable_flag', $time);
            }`
I don't think this check is required. Instead of that if we use `update_option` then it will add the flag if does not exist. Refer this [link](https://developer.wordpress.org/reference/functions/update_option/)

jampulanaveen: ok i will update it

nikhilsharma1100 (approved):
LGTM

PR #282, opened 2022-05-23
bugfix

jampulanaveen:


includes/razorpay-webhook.php
```diff
@@ -76,15 +76,13 @@ public function process()
...
-      
-       
         if (empty($data['event']) === false) {
             // Skip the webhook if not the valid data and event
             if ($this->shouldConsumeWebhook($data) === false) {
                 return;
             }
             if (isset($_SERVER['HTTP_X_RAZORPAY_SIGNATURE']) === true) {
-                $razorpayWebhookSecret = get_option('rzp_webhook_secret');
+                $razorpayWebhookSecret = $this->razorpay->getSetting('webhook_secret');
```
ChetanGN, line 85: @jampulanaveen We need to handle the case where merchants were onboarded with the 3.7.1 version, where secret will be stored with `rzp_webhook_secret` key.

ChetanGN (approved):
LGTM

PR #314, opened 2022-07-14
Bug fix in logs

ChetanGN:


woo-razorpay.php
```diff
@@ -1342,7 +1342,7 @@ public function updateOrder(& $order, $success, $errorMessage, $razorpayPaymentI
 
             $orderId = $order->get_order_number();
 
-            rzpLogInfo("updateOrder orderId: $orderId , errorMessage: $errorMessage, razorpayPaymentId: $razorpayPaymentId , success: $success");
```
abdulwahidsharief, line 1345: @ChetanGN why $errorMessage is removed ?

ChetanGN: It was leading to confusion as merchant logs have this error message even if the payment was successful. So we don't need to print this on logs. @abdulwahidsharief

poojarazor (approved):
LGTM

angbpy (approved):
LGTM

PR #325, opened 2022-08-03
[PLG-426]fix duplicate wooorderid and change amt validation

poojarazor:


woo-razorpay.php
```diff
@@ -818,11 +818,16 @@ protected function verifyOrderAmount($razorpayOrderId, $orderId)
 
             $razorpayOrderArgs = array(
                 'id'        => $razorpayOrderId,
-                'amount'    => $orderCreationData['amount'],
                 'currency'  => $orderCreationData['currency'],
                 'receipt'   => (string) $orderId,
             );
 
+            if(is1ccEnabled()){
```
ChetanGN, line 825: @poojarazor don't use this flag for condition, instead of that use `is1ccCheckout` by passing in params.

woo-razorpay.php
```diff
@@ -796,7 +796,7 @@ protected function createRazorpayOrderId($orderId, $sessionKey)
             return $razorpayOrderId;
         }
 
-        protected function verifyOrderAmount($razorpayOrderId, $orderId)
+        protected function verifyOrderAmount($razorpayOrderId, $orderId, $is1ccCheckout)
```
ChetanGN, line 799: You can pass the default value for `$is1ccCheckout` as no here.

ChetanGN (approved):
LGTM. Test it out after merging with master

angbpy (approved):

PR #332, opened 2022-08-12
Woodmart Sticky AddToCart Support

angbpy:


btn-1cc-checkout.js
```diff
@@ -43,6 +43,56 @@ function btnCheckout(){
     });
   }
 
+  // For attaching event listener to Woodmart's sticky add-to-cart
+  document.addEventListener('scroll',(e)=>{
+     
+    var stickyBtn=document.querySelectorAll('#btn-1cc-pdp')[1];
+
+    if (stickyBtn != null) {
+        stickyBtn.onclick = function() {
```
ChetanGN, line 52: Can we make it a common function and invoke it in both places? @angbpy

angbpy: @ChetanGN Done!

ChetanGN:
@angbpy Quantity is not updating through the sticky header. Please fix it.

poojarazor (approved):
LGTM

btn-1cc-checkout.js
```diff
@@ -43,47 +43,153 @@ function btnCheckout(){
     });
   }
 
+  // For attaching event listener to Woodmart's sticky add-to-cart
+  document.addEventListener('scroll',(e)=>{
+     
+    var stickyBtn=document.querySelectorAll('#btn-1cc-pdp')[1];
```
ChetanGN, line 49: @angbpy This var will be null for all other pages except the product display page. Please handle this case

PR #338, opened 2022-08-26
Cartbounty bug fix for variant products

angbpy:


ChetanGN:
@angbpy we should not invoke `handleCBRecoveredOrder` in case of native checkout. Please add the condition on the callback function.

angbpy:
> @angbpy we should not invoke `handleCBRecoveredOrder` in case of native checkout. Please add the condition on the callback function.

Updating by checking the 1cc condition

woo-razorpay.php
```diff
@@ -1552,7 +1552,9 @@ public function updateOrder(& $order, $success, $errorMessage, $razorpayPaymentI
                     $order->payment_complete($razorpayPaymentId);
                 }
 
-                if(is_plugin_active('woo-save-abandoned-carts/cartbounty-abandoned-carts.php')){
+                $is1ccOrder = get_post_meta( $wcOrderId, 'is_magic_checkout_order', true );
```
ChetanGN, line 1555: @angbpy Why do we need to get the post meta info here?

angbpy: We are receiving the info whether the order that is placed is a 1cc order or not from the post meta table

ChetanGN: @angbpy It's already defined in this [line](https://github.com/razorpay/razorpay-woocommerce/blob/736fbb04c7753c01ec503e77a9b6224bba7814a4/woo-razorpay.php#L1521), can we utilize the same?

angbpy: In case of an error in the try block will it still reach this line or then we shouldn't worry about it?

ChetanGN (approved):

PR #343, opened 2022-09-02
fix for recursive redirect issue

enishant:
Recursive redirect issue fix

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -233,8 +233,6 @@ function woocommerce_razorpay_init()
 
         protected function initHooks()
         {
-            add_action('init', array(&$this, 'check_razorpay_response'));
-
             add_action('woocommerce_receipt_' . $this->id, array($this, 'receipt_page'));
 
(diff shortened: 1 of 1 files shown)


woo-razorpay.php
```diff
@@ -233,8 +233,6 @@ public function __construct($hooks = true)
 
         protected function initHooks()
         {
-            add_action('init', array(&$this, 'check_razorpay_response'));
```
ChetanGN, line 236: @enishant Can we test these changes with VAS merchant keys?

enishant: Tested standard flow and VAS checkout with latest WordPress + WooCommerce version,  payment flow and callback redirection working fine.

While executing vas checkout, JS script was not loading in document. For testing changed script inclusion as follow to include JS script after DOM loaded within document.

`wp_enqueue_script('razorpay_wc_script');`
**Changed to following**
`wp_enqueue_script('razorpay_wc_script',false, [], false, true );`

@ChetanGN @abdulwahidsharief 
 Please check if we need to include this change as script is executing before DOM loaded.

ChetanGN (approved):

PR #346, opened 2022-09-06
fixed Order properties should not be accessed directly.

abdulwahidsharief:
fixed . Order properties should not be accessed directly and undefined variable $name


poojarazor (approved):
LGTM

PR #352, opened 2022-09-09
push plugin events to datalake

abdulwahidsharief:
datalake changes for woocommerce to push plugin events.

### DIFF
--- a/includes/plugin-instrumentation.php
+++ b/includes/plugin-instrumentation.php
@@ -5,11 +5,13 @@ use Razorpay\Api\Errors;
 
 class TrackPluginInstrumentation
 {
-    public $api;
+    protected $api;
+    protected $mode;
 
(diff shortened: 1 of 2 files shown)


includes/plugin-instrumentation.php
```diff
@@ -65,19 +106,25 @@ public function rzpTrackDataLake($properties)
         }
     }
 
-    public function getDefaultProperties()
+    public function getDefaultProperties($timestamp = true)
```
enishant, line 109: @abdulwahidsharief timestamp is always required, it's not optional.

abdulwahidsharief: @enishant in datalake we have timestamp as required field in schema it will be data redundancy if we keep same timestamp in properties

enishant (approved):
LGTM

PR #373, opened 2022-10-19
Abandoned plugin support ( Action Hook)

angbpy:


includes/api/save-abandonment-data.php
```diff
@@ -59,6 +60,8 @@ function saveCartAbandonmentData(WP_REST_Request $request)
     $result['response']    = "";
     $result['status_code'] = 400;
 
+    abandonedPluginHook($razorpayData); // do_action to notify/send the abandonedCart data to 3rd party plugins
```
ChetanGN, line 63: This function should call right after the try-catch block if there is no other dependency.

includes/support/abandoned-cart-hooks.php
```diff
@@ -0,0 +1,53 @@
...
+     $rzpAbandonedData                                = array();
+     $rzpAbandonedData['woocommerceOrderID']          = $razorpayData['receipt'];
+     $rzpAbandonedData['amount']                      = $razorpayData['amount'];
+     $rzpAbandonedData['currency']                    = $razorpayData['currency'];
+     $rzpAbandonedData['customer_details']            = array();
+     $rzpAbandonedData['customer_details']['contact'] = $razorpayData['customer_details']['contact'];
+     $rzpAbandonedData['customer_details']['email']   = $razorpayData['customer_details']['email'];
+
+    $shippingAddress                                  = array(
+        'name'     => $razorpayData['customer_details']['shipping_address']['name'],
```
ChetanGN, line 14: Define the variable with the shipping address like
`$magicShippingAddress = $razorpayData['customer_details']['shipping_address']` and use this here. Same for the billing address.

includes/support/abandoned-cart-hooks.php
```diff
@@ -0,0 +1,53 @@
...
+
+function abandonedPluginHook($razorpayData) {
+
+     $rzpAbandonedData                                = array();
+     $rzpAbandonedData['woocommerceOrderID']          = $razorpayData['receipt'];
+     $rzpAbandonedData['amount']                      = $razorpayData['amount'];
+     $rzpAbandonedData['currency']                    = $razorpayData['currency'];
+     $rzpAbandonedData['customer_details']            = array();
+     $rzpAbandonedData['customer_details']['contact'] = $razorpayData['customer_details']['contact'];
+     $rzpAbandonedData['customer_details']['email']   = $razorpayData['customer_details']['email'];
```
ChetanGN, line 11: As email will be an optional field in the modal, will that affect the flow? varify this @angbpy

ChetanGN (changes requested):

includes/support/abandoned-cart-hooks.php
angbpy: This shouldn't affect the flow as email empty/not-empty checks are performed at the 3rd party plugin end before insertion

ChetanGN: What if the email key does not exist in rzp object? verify this flow.

poojarazor (approved):
LGTM

angbpy (commented):
Changes already made

ChetanGN (approved):

PR #374, opened 2022-11-03
Mini cart + Electro Mobile button + Spinner complete fix

angbpy:


poojarazor (approved):
Before merge test it with multiple theme

PR #388, opened 2022-12-18
disable coupon with dynamic price

poojarazor:


angbpy (approved):

includes/api/order.php
```diff
@@ -119,8 +119,31 @@ function createWcOrder(WP_REST_Request $request)
...
+        // Woo dynamic discount price plugin
+        if(is_plugin_active('yith-woocommerce-dynamic-pricing-and-discounts-premium/init.php')) {
+
+            foreach ($order->get_items() as $itemId => $item) {
+                $dynamicRules = $item->get_meta('_ywdpd_discounts');
+
+                foreach ($dynamicRules['applied_discounts'] as $appliedDiscount) {
+                    if (isset( $appliedDiscount['set_id'])){
+                        $ruleId = $appliedDiscount['set_id'];
+                        $rule    = ywdpd_get_rule($ruleId);
```
ChetanGN, line 136: @poojarazor Do we need to initiate a class? to invoke this function?

includes/api/order.php
```diff
@@ -216,6 +239,10 @@ function createWcOrder(WP_REST_Request $request)
         $response['redirect']            = true;
         $response['one_click_checkout']  = true;
 
+        if($disableCouponFlag == true){
+           $response['show_coupons']  = false;
```
ChetanGN, line 243: Do we need to pass default value?

includes/api/order.php
```diff
@@ -119,8 +119,31 @@ function createWcOrder(WP_REST_Request $request)
...
-    if ($order) {
+    
+    if($order){
+
+        $disableCouponFlag = false;
+
+        // Woo dynamic discount price plugin
+        if(is_plugin_active('yith-woocommerce-dynamic-pricing-and-discounts-premium/init.php')) {
+
+            foreach ($order->get_items() as $itemId => $item) {
```
ChetanGN, line 130: Can we get the TCD's for this PR?

includes/api/order.php
poojarazor: it's a public function , no need to initiate any class

includes/api/order.php
poojarazor: no required

includes/api/order.php
poojarazor: Discussed with Gaurav he said  already test case is written for this plugin so no separate TCD required , only I have to do feature testing from my side.

PR #389, opened 2022-12-22
condtional and shipping plugin coupon restriction

poojarazor:


includes/api/shipping-info.php
```diff
@@ -593,6 +598,54 @@ function smartCodRestriction($addresses, $order)
...
+    try
+    {
+        $razorpayData = $api->order->fetch($rzpOrderId);
+
+    } catch (Exception $e) {
+
+        return true;
+    }
+
+    // foreach($razorpayData['promotions'] as $promotion)
```
ChetanGN, line 617: Why this condition is commented?

poojarazor: The GC code was not merged at that time, so I commented it , now removed the commented code

angbpy (approved):

PR #436, opened 2023-02-28
remove is admin code

poojarazor:

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -171,7 +171,7 @@ function woocommerce_razorpay_init()
             $isAccCreationAvailable = false;
 
             // Load preference API call only for administrative interface page.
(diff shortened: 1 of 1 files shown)


woo-razorpay.php
```diff
@@ -171,7 +171,7 @@ public function __construct($hooks = true)
             $isAccCreationAvailable = false;
 
             // Load preference API call only for administrative interface page.
-            if (is_admin())
+            if (current_user_can('administrator'))
```
ChetanGN, line 174: How this makes a difference? @poojarazor

poojarazor: Above condition is not working properly @ChetanGN

ChetanGN (approved):
LGTM

PR #444, opened 2023-03-17
Removed woocommerce-message class to fix redirect issue in Elessi theme

sonyy-m:

Html content of 'woocommerce-message' class is getting replaced to some other text message in Elessi theme.
https://razorpay.atlassian.net/browse/PLUG-1611
Removing this class to fix the issue.


abdulwahidsharief:
LGTM

abdulwahidsharief (approved):

PR #456, opened 2023-04-19
fix: removed types of variable

Anantha-Perumal:

Fix: removed type of variables


angbpy (approved):
LGTM

ChetanGN (approved):
LGTM

PR #459, opened 2023-05-04
One cc cron db update fix: php v8.2

Anantha-Perumal:


angbpy:
@Anantha-Perumal Can you explain the changes here ?
Why it was breaking and what fixed it

Anantha-Perumal:
@angbpy We are fetching data from DB by `get_option` if data is not there it will return false but we are accessing the value as an array. Hence while accessing the boolean value typed variable as an array in PHP 8.2, it throws a warning hence making the variable an array if data is not present. For exact warning message please refer [here](https://razorpay.slack.com/archives/C0132U6BQBY/p1683135092891329?thread_ts=1683135079.419409&cid=C0132U6BQBY)

poojarazor (approved):
LGTM

PR #471, opened 2023-06-27
Ut for Razorpay Route

garimaprachi:


includes/razorpay-route.php
```diff
@@ -1262,4 +1267,4 @@ function razorpayPaymentTransfer()
     $routeAction = new RZP_Route_Action();
 
     $routeAction->createPaymentTransfer();
-}
+}
```
abdulwahidsharief, line 1270: @garimaprachi add EOL

abdulwahidsharief (approved):
LGTM

PR #480, opened 2023-07-06
Ut for razorpay route

garimaprachi:


tests/phpunit/tests/test-razorpay-route-actions.php
```diff
@@ -16,6 +17,171 @@ public function setup(): void
...
+
+        $pageUrl = admin_url('admin.php?page=razorpayTransfers&id=' . 'test');
+        $this->instance->shouldReceive('redirect')->with($pageUrl);
+
+        $response = $this->instance->reverseTransfer();
+
+        $this->assertNull($response);
+    }
+
+    public function testDirectTransfer()
```
abdulwahidsharief, line 39: @garimaprachi add negative / exception block test cases

garimaprachi: A wp_die statement is there so the function will die because of which phpunit wont send in result
Hence it is not possible.

abdulwahidsharief (approved):
LGTM

PR #483, opened 2023-07-20
smart cod issue fix

poojarazor:

### DIFF
--- a/includes/api/shipping-info.php
+++ b/includes/api/shipping-info.php
@@ -640,7 +640,7 @@ function smartCodRestriction($addresses, $order)
 
 function codTrim($str){
 
-  $str=strtoupper($str);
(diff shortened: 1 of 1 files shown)


ChetanGN (approved):
LGTM

includes/api/shipping-info.php
```diff
@@ -638,6 +638,17 @@ function smartCodRestriction($addresses, $order)
     return true;
 }
 
+function codTrim($str){
+
+  $str=strtoupper($str);
```
ChetanGN, line 643: The function name tells as it's job is trimming, but why are we setting it to upper case? @poojarazor

poojarazor: replaced by trim

PR #484, opened 2023-07-21
Updated webhook secret fetch when setting is empty.

yashgit891:

### DIFF
--- a/includes/razorpay-webhook.php
+++ b/includes/razorpay-webhook.php
@@ -100,7 +100,9 @@ class RZP_Webhook
                 return;
             }
             if (isset($_SERVER['HTTP_X_RAZORPAY_SIGNATURE']) === true) {
-                $razorpayWebhookSecret = $this->razorpay->getSetting('webhook_secret');
+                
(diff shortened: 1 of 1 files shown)


includes/razorpay-webhook.php
```diff
@@ -100,7 +100,9 @@ public function process()
                 return;
             }
             if (isset($_SERVER['HTTP_X_RAZORPAY_SIGNATURE']) === true) {
-                $razorpayWebhookSecret = $this->razorpay->getSetting('webhook_secret');
+                
+                $razorpayWebhookSecret = (empty($this->razorpay->getSetting('webhook_secret')) === false) ? $this->razorpay->getSetting('webhook_secret') : get_option('webhook_secret');
```
abdulwahidsharief, line 104: @yashgit891 do you have idea why there is another webhook secret rzp_webhook_secret flag

yashgit891: @abdulwahidsharief it was present for few of the older merchants before the flag was renamed to webhook_secret.

abdulwahidsharief (approved):
LGTM

PR #486, opened 2023-07-27
Hpos instrumentation

yashgit891:

### DIFF
--- a/includes/plugin-instrumentation.php
+++ b/includes/plugin-instrumentation.php
@@ -53,7 +53,6 @@ class TrackPluginInstrumentation
             update_option('rzp_hpos', 'yes');
         }
         else if(OrderUtil::custom_orders_table_usage_is_enabled() === false and
-                empty(get_option('rzp_hpos')) === false and
                 get_option('rzp_hpos') === 'yes')
         {
             $properties = [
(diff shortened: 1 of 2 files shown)


includes/plugin-instrumentation.php
```diff
@@ -30,11 +30,43 @@ function razorpayPluginActivated()
...
+            $properties = [
+                'isHposEnabled' => true
+            ];
+
+            $response = $this->rzpTrackSegment('hpos.interacted', $properties);
+            $this->rzpTrackDataLake('hpos.interacted', $properties);
+
+            update_option('rzp_hpos', 'yes');
+        }
+        else if(OrderUtil::custom_orders_table_usage_is_enabled() === false and
```
abdulwahidsharief, line 55: @yashgit891 can we use else here

includes/plugin-instrumentation.php
```diff
@@ -30,11 +30,43 @@ function razorpayPluginActivated()
...
+                empty(get_option('rzp_hpos')) === false and
```
abdulwahidsharief, line 56: get_option('rzp_hpos') === 'yes' condition will do job of empty(get_option('rzp_hpos')) === false. a condition can be removed

includes/plugin-instrumentation.php
yashgit891: @abdulwahidsharief no as it would lead to false positives.

abdulwahidsharief (approved):
LGTM

PR #495, opened 2023-08-24
Order status fix

poojarazor:

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -1651,6 +1651,8 @@ EOT;
                         $order->set_status('wc-pending');
                         $order->save();
                     }
+                    $orderStat = $order->get_status();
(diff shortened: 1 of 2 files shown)


woo-razorpay.php
```diff
@@ -1684,10 +1688,12 @@ function check_razorpay_response()
 
             // If the order has already been paid for
             // redirect user to success page
-            if ($order->needs_payment() === false)
+            if (($order->needs_payment() === false) and 
+                ($orderStat != 'draft') and 
+                ($orderStat != 'wc-checkout-draft'))
```
abdulwahidsharief, line 1693: @poojarazor can we use in_array() here

poojarazor: @abdulwahidsharief  don't want to initialise the array for the statement

abdulwahidsharief (approved):
LGTM

PR #503, opened 2023-09-11
new sdk 2.8.7 changes

poojarazor:


ankitdas13 (approved):
Before merge, please check other api's (if woocommerce is using other api rather than `order create`) as well, because now in sdk we are handling `v1` & `v2` api version both. 

LGTM

abdulwahidsharief (approved):
LGTM

PR #513, opened 2023-11-21
added authentication and authorization for route

abdulwahidsharief:


ramth05 (approved):
LGTM, please test it properly in kinsta stage site, before merging.

PR #526, opened 2024-01-17
TerraWallet

Anantha-Perumal:


includes/api/cart.php
```diff
@@ -39,6 +39,10 @@ function fetchCartData(WP_REST_Request $request)
 
     $response = cartResponse($couponCode);
 
+    $response['user'] = getCartUserObject();
```
ChetanGN, line 42: Why do we need this info if the Terra Wallet plugin is not activated on the store? Will it not add some latency to cart fetch/create API?

woo-razorpay.php
```diff
@@ -1496,7 +1496,7 @@ public function process_refund($orderId, $amount = null, $reason = '')
         }
 
         // process refund for gift card
-        function processGiftCardRefund($orderId, $razorpayPaymentId, $amount = null, $reason = '')
+        function processRefund($orderId, $razorpayPaymentId, $amount = null, $reason = '')
```
ChetanGN, line 1499: Already function named with `process_refund` is present. Can we reuse that or give a different name which specify it's related to magic-related for terra-wallet!

woo-razorpay.php
```diff
@@ -2274,21 +2274,53 @@ public function updateGiftAndCoupon($razorpayData, $order, $orderId, $razorpayPa
...
+                }else if($promotion['type'] == 'terra_wallet') {
+                    if (!is_plugin_active('woo-wallet/woo-wallet.php')) {
+                        $this->processRefund(
+                            $orderId,
+                            $razorpayPaymentId,
+                            $razorpayData['amount_paid'],
+                            'Terra Wallet Plugin is not active');
+                    }
+                    $viaWalletAmount = $promotion['value']/100;
+                    $walletBalance = woo_wallet()->wallet->get_wallet_balance($order->get_user_id(), 'edit');
```
ChetanGN, line 2300: Where this `woo_wallet` class initiated? If plugin is not active or uninstalled then it will though fatal error right?

woo-razorpay.php
```diff
@@ -2274,21 +2274,53 @@ public function updateGiftAndCoupon($razorpayData, $order, $orderId, $razorpayPa
...
+                    if ($walletBalance < $viaWalletAmount)
```
ChetanGN, line 2301: Need to understand the logic.

includes/api/cart.php
Anantha-Perumal: @ChetanGN we need this to fetch store's logged in user email

woo-razorpay.php
Anantha-Perumal: sure

woo-razorpay.php
Anantha-Perumal: no in line no 2292 i am checking whether plugin is active or not

woo-razorpay.php
Anantha-Perumal: if the walletbalance(amount available in terrawallet) is less than viaWalletAmount(amount redeemed for an order) then i am processing refund as there is insufficient balance

ChetanGN (approved):
LGTM

PR #530, opened 2024-01-23
Instant refund speed addition

yashgit891:

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -1465,6 +1465,7 @@ EOT;
 
             $data = array(
                 'amount'    =>  (int) round($amount * 100),
+                'speed'     => "optimum",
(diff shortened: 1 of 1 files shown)


woo-razorpay.php
```diff
@@ -1487,6 +1488,10 @@ public function process_refund($orderId, $amount = null, $reason = '')
                  */
                 do_action( 'woo_razorpay_refund_success', $refund->id, $orderId, $refund );
 
+                rzpLogInfo('Refund ID' . $refund->id);
```
ramth05, line 1491: @yashgit891 , please have checks for object available before printing or using it.
Also make in array use all 3 logs print in one call.

abdulwahidsharief (approved):
LGTM

PR #538, opened 2024-02-23
Shipping Engine Support for woocommerce

Anantha-Perumal:
This PR sends product ID, variant ID and weight of the product while order creation


ChetanGN (approved):
LGTM

PR #539, opened 2024-02-26
[PLUGIN1CC - 3282] adding shipping method name for shipping engine

nkttiwari:

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -2150,6 +2150,10 @@ EOT;
                 if($razorpayData['shipping_fee'] == 0)
                 {
                     $item->set_method_title( 'Free Shipping' );
+                    if(isset($razorpayData['notes']) && isset($razorpayData['notes']['shipping_method_name']))
+                    {
+                        $item->set_method_title($razorpayData['notes']['shipping_method_name']);
+                    }
                 }
                 else
                 {
@@ -2221,9 +2225,9 @@ EOT;
                         // $item->set_method_id( "flat_rate:1" );
                         $item->set_total( $razorpayData['shipping_fee']/100 );
 
-                        if (isset($razorpayData['shipping_method']))
+                        if(isset($razorpayData['notes']) && isset($razorpayData['notes']['shipping_method_name']))
                         {
-                            $item->set_method_title($razorpayData['shipping_method']['name']);
+                            $item->set_method_title($razorpayData['notes']['shipping_method_name']);
                         }
 
                         $order->add_item( $item );
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -2221,6 +2221,11 @@ EOT;
                         // $item->set_method_id( "flat_rate:1" );
                         $item->set_total( $razorpayData['shipping_fee']/100 );
 
+                        if (isset($razorpayData['shipping_method']))
+                        {
+                            $item->set_method_title($razorpayData['shipping_method']['name']);
+                        }
+
                         $order->add_item( $item );
 
                         $item->save();


woo-razorpay.php
```diff
@@ -2215,6 +2215,11 @@ public function update1ccOrderWC(& $order, $wcOrderId, $razorpayPaymentId)
                         // $item->set_method_id( "flat_rate:1" );
                         $item->set_total( $razorpayData['shipping_fee']/100 );
 
+                        if (isset($razorpayData['shipping_method']))
```
Anantha-Perumal, line 2218: this change is in else block, when will the if block execute? 
there also we are adding WC_Order_Item_shipping

nkttiwari: the if block execute for wcfm plugins shipping for woo Commerce, doesn't think that this changes should be added there too. will discuss with @ChetanGN on the same.

Anantha-Perumal: please check with chetan and refactor your condition to `isset($razorpayData['shipping_method']) &&  isset($razorpayData['shipping_method']['name'])`

Anantha-Perumal: also we need to set shipping method name for 0 shipping fee as well

woo-razorpay.php
```diff
@@ -2150,6 +2150,10 @@ public function update1ccOrderWC(& $order, $wcOrderId, $razorpayPaymentId)
                 if($razorpayData['shipping_fee'] == 0)
                 {
                     $item->set_method_title( 'Free Shipping' );
+                    if(isset($razorpayData['notes']) && isset($razorpayData['notes']['shipping_method_name']))
```
ChetanGN, line 2153: `notes` key will be always present so I don't think we need to check `isset($razorpayData['notes'])`

woo-razorpay.php
```diff
@@ -2150,6 +2150,10 @@ public function update1ccOrderWC(& $order, $wcOrderId, $razorpayPaymentId)
...
+                    {
+                        $item->set_method_title($razorpayData['notes']['shipping_method_name']);
```
ChetanGN, line 2155: Can we add a default shipping name if that key is not available? Like bellow
`$shippingMethodTitle = $razorpayData['notes']['shipping_method_name'] ?? 'Standard Shipping';
$item->set_method_title($shippingMethodTitle);`

Anantha-Perumal: @nkttiwari what happens if shipping_method_name is empty?

woo-razorpay.php
```diff
@@ -2221,6 +2225,11 @@ public function update1ccOrderWC(& $order, $wcOrderId, $razorpayPaymentId)
                         // $item->set_method_id( "flat_rate:1" );
                         $item->set_total( $razorpayData['shipping_fee']/100 );
 
+                        if(isset($razorpayData['notes']) && isset($razorpayData['notes']['shipping_method_name']))
+                        {
+                            $item->set_method_title($razorpayData['notes']['shipping_method_name']);
+                        }
+
```
Anantha-Perumal, line 2232: @nkttiwari we haven't handled the case where storeShippingIsEnabled at (_line no: 2186_) can you please check if this case come into play with our shipping engine?

i see checking whether shipping_method_name is duplicated in two places can you please fetch shipping method and assign it to a variable and use that wherever it is necessary.
And because of you changes, line no from 2215-2222 and 2152 will not have any effect if order notes have shipping_method_name so handle this case accordingly -> fetch method title at a common place and set title in a single place.

ChetanGN (approved):

PR #541, opened 2024-03-08
fix: converting weight to grams based on wooc weight metric

Anantha-Perumal:
### Issue
COD & Shipping fee evaluation are done in a wrong way for woocommerce Merchant who are all having their weight metrics other than kg

### Root Cause
Weight is directly converted to grams with the assumption that it is always in KG.

### Fix
Fetched the store's weight metric configuration and converting the weight to grams based on the metric.

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -1383,7 +1383,7 @@ function woocommerce_razorpay_init()
                    $product_id = $product->get_parent_id();
                }
                $data['line_items'][$i]['product_id'] = (string)$product_id;
-			   $data['line_items'][$i]['weight'] = $product->get_weight()*1000*(int)$item->get_quantity();
+			   $data['line_items'][$i]['weight'] = $this->getWeightBasedOnStoreSettingsAndQuantity($product->get_weight(), $item->get_quantity());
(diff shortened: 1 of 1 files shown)


woo-razorpay.php
```diff
@@ -1383,7 +1383,7 @@ public function orderArg1CC($data, $order)
                    $product_id = $product->get_parent_id();
                }
                $data['line_items'][$i]['product_id'] = (string)$product_id;
-			   $data['line_items'][$i]['weight'] = $product->get_weight()*1000*(int)$item->get_quantity();
+			   $data['line_items'][$i]['weight'] = $this->getWeightBasedOnStoreSettingsAndQuantity($product->get_weight(), $item->get_quantity());
```
ChetanGN, line 1386: @Anantha-Perumal Have you tried [this](https://wp-kama.com/plugin/woocommerce/function/wc_get_weight) wooc default method which matches with our requirement? 
If this function is not working then we can go ahead with the custom function which you had written.

Anantha-Perumal: tried it, it works refactored the code

ChetanGN (approved):

PR #546, opened 2024-03-27
Added null check in upgrade instrumentation

yashgit891:


public/js/admin-rzp-settings.js
```diff
@@ -41,6 +41,11 @@ window.onload = function() {
 	{
 		var rzpElement = document.getElementById(item);
 
+		if(rzpElement == null)
```
abdulwahidsharief, line 44: @yashgit891 leave a whitespace after if

abdulwahidsharief (approved):
LGTM

PR #548, opened 2024-04-25
[PLUGIN1CC-3639] Added magic checkout admin banner on wooc dashboard

ChetanGN:


woo-razorpay.php
```diff
@@ -335,6 +335,8 @@ protected function initHooks()
         {
             add_action('woocommerce_receipt_' . $this->id, array($this, 'receipt_page'));
 
+            add_action( 'admin_notices', array($this, 'add_magic_banner' ));
```
abdulwahidsharief, line 338: @ChetanGN remove whitespave before and after common brackets

PR #563, opened 2024-06-13
PO-136 remove default currency as INR

abdulwahidsharief:

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -1127,7 +1127,7 @@ function woocommerce_razorpay_init()
             return array(
                 'key'          => $this->getSetting('key_id'),
                 'name'         => html_entity_decode(get_bloginfo('name'), ENT_QUOTES),
-                'currency'     => $this->getOrderCurrency($order),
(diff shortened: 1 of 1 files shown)


woo-razorpay.php
```diff
@@ -1127,7 +1127,7 @@ public function getDefaultCheckoutArguments($order)
             return array(
                 'key'          => $this->getSetting('key_id'),
                 'name'         => html_entity_decode(get_bloginfo('name'), ENT_QUOTES),
-                'currency'     => self::INR,
+                'currency'     => $this->getOrderCurrency($order),
```
yashgit891, line 1130: @abdulwahidsharief can we assign this value to a variable and then pass it to the array, just like it has been done for other values?

abdulwahidsharief: https://github.com/razorpay/razorpay-woocommerce/pull/563/commits/4e648e38ef9c6697fdb5baacc57f4b83a801bab9

rohitcbr (approved):

PR #578, opened 2024-12-06
PLUG-2874 woocommerce checkout 360

abdulwahidsharief:

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -248,6 +248,166 @@ function woocommerce_razorpay_init()
             return $message;
         }
 
+        public function generate_multicheckbox_html($key, $data)
+        {
+            $c360FieldMapping = [
+                'enable_magic_checkout_cart_page'           => 'enable_1cc',
+                'enable_magic_checkout_test_mode'           => 'enable_1cc_test_mode',
+                'enable_magic_checkout_buy_now'             => 'enable_1cc_pdp_checkout',
+                'enable_magic_checkout_mini_cart'           => 'enable_1cc_mini_cart_checkout',
+                'enable_cod_intelligence'                   => 'rzp_cod_intelligence_enable',
+                'enable_razorpay_trusted_business_widget'   => 'rzp_rtb_enable',
+                'enable_razorpay_affordability_widget'      => 'rzp_afd_enable'
+            ];
+
+            $field = $this->get_field_key($key);
+
+            $defaults = array(
+                'title'       => '',
+                'type'        => 'multicheckbox',
+                'description' => '',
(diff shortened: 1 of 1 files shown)


woo-razorpay.php
```diff
@@ -248,14 +248,199 @@ public function getCustomOrdercreationMessage($thank_you_title, $order)
...
+        public function toggleRtbWidget($action ,$widgetEnabled = false)
+        {
+            $data = [
+                'rtb_widget_enabled' => $widgetEnabled,
+            ];
+
+            $rtbEnable = $this->rzpPostInstallationCall('save', $data);
+
+            // Experiment returned false
+            if (isset($rtbEnable['checkout360_status']) === true and
```
rohitcbr, line 320: @abdulwahidsharief since this api call(rzpPostInstallationCall) may return error in case of exception(return ['error' => $e->getMessage()]), can we put a check for the error first?

woo-razorpay.php
```diff
@@ -535,6 +842,43 @@ protected function triggerValidationInstrumentation($data)
             $trackObject->rzpTrackDataLake('formfield.validation.error', $properties);
         }
 
+        public function autoPostInstallationCheck()
+        {
+            $response = $this->rzpPostInstallationCall('evaluate');
```
rohitcbr, line 847: @abdulwahidsharief can we put an error check after this call?

woo-razorpay.php
```diff
@@ -535,6 +842,43 @@ protected function triggerValidationInstrumentation($data)
...
+
+            $checkout360Available = ($response['checkout360_status'] === true) ? 'yes' : 'no';
```
rohitcbr, line 849: @abdulwahidsharief should we put an `isset` check too?

woo-razorpay.php
```diff
@@ -535,6 +842,43 @@ protected function triggerValidationInstrumentation($data)
...
+
+            update_option('rzp_checkout360_status', $checkout360Available);
+
+            // remove all checkou360 settings
+            if ($response['merchant_status'] === 'existing_checkout360_merchant' and
```
rohitcbr, line 854: @abdulwahidsharief should we put an `isset` check here?

woo-razorpay.php
abdulwahidsharief: @rohitcbr as error is logged in rzpPostInstallationCall just keeping a check for error to return will increase processing time for 200 response requests

rohitcbr (approved):

rohitcbr:
LGTM, please test all the scenarios after addressing the review comments

PR #579, opened 2024-12-11
["PO-252"]("WooCommerce : add isset validation for webhook payload")

rohitcbr:

### DIFF
--- a/includes/razorpay-webhook.php
+++ b/includes/razorpay-webhook.php
@@ -526,9 +526,8 @@ class RZP_Webhook
     protected function shouldConsumeWebhook($data)
     {
         if ((isset($data['event']) === true) and
-            ((in_array($data['event'], $this->eventsArray) === true) or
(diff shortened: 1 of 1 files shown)


includes/razorpay-webhook.php
```diff
@@ -96,6 +96,13 @@ public function process()
             return;
         }
 
+        // Skip the webhook if not the valid data and event
+        if ($this->shouldConsumeWebhook($data) === false) {
+
+            rzpLogDebug("Webhook process exited in shouldConsumeWebhook function");
```
abdulwahidsharief, line 102: @rohitcbr please change message and log data as you are using  rzpLogDebug

rohitcbr: @abdulwahidsharief updated here https://github.com/razorpay/razorpay-woocommerce/pull/579/commits/0ee45d4010556d22d68a9a8a0f263b6a38c8d8b6

abdulwahidsharief (approved):
LGTM

PR #619, opened 2025-08-13
ISS-973762 fix magic checkout enqueue scripts

abdulwahidsharief:

### DIFF
--- a/woo-razorpay.php
+++ b/woo-razorpay.php
@@ -2713,6 +2713,19 @@ EOT;
 
                 rzpLogInfo("updateOrderAddress function executed");
 
+                $email = $razorpayData['customer_details']['email'];
+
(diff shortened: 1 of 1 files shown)


woo-razorpay.php
```diff
@@ -3509,7 +3510,6 @@ function disable_coupon_field_on_cart($enabled)
 if(is1ccEnabled())
 {
     add_filter('woocommerce_coupons_enabled', 'disable_coupon_field_on_cart');
-    add_action('woocommerce_cart_updated', 'enqueueScriptsFor1cc', 10);
```
ChetanGN, line 3512: @abdulwahidsharief As we removed this hook, can you please verify the 1cc buttons are properly rendering on cart update as well?

PR #625, opened 2025-09-15
fixed validation for woocommerce order id if returned zero

abdulwahidsharief:

### DIFF
--- a/includes/api/order.php
+++ b/includes/api/order.php
@@ -81,7 +81,7 @@ function createWcOrder(WP_REST_Request $request)
             $orderId  = $checkout->create_order(array());
 
             if (is_wp_error($orderId) || empty($orderId)) {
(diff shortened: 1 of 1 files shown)


includes/api/order.php
```diff
@@ -80,8 +80,8 @@ function createWcOrder(WP_REST_Request $request)
             $checkout = WC()->checkout();
             $orderId  = $checkout->create_order(array());
 
-            if (is_wp_error($orderId)) {
-                $checkout_error = $orderId->get_error_message();
+            if (is_wp_error($orderId) || empty($orderId)) {
+                $checkout_error = is_wp_error($orderId) ? $orderId->get_error_message() : "Invalid order ID returned:". $orderId;
```
amisha25-razor, line 84: We already have a condition for handling empty $orderId. In that case, do we really need to append the $orderId in the error message?
If $orderId is empty, this will result in just "Invalid order ID returned:" (with nothing after the colon).
Do we really need to pass $orderId in the  error message?

PR #647, opened 2026-04-08
[Security] Fix prepayCODOrder - correct meta key lookup and signature verification

gutti-yaswanth:
Validate Razorpay order ID against stored WooCommerce order meta
Add protection against duplicate payment processing
Add HMAC-based authentication using checkHmacSignature

## Note :- Please follow the below points while attaching test cases document link below:
   ### - If label `Tested` is added then test cases document URL is mandatory.

## Summary                                                                                                                                                                                                
                                                                                                                                                                                                            
  Fixes CVE-2026-3507 — an unauthenticated IDOR vulnerability in the                                                                                                                                        
  `prepayCODOrder` endpoint (`/wp-json/1cc/v1/cod/order/prepay`) that allowed
  attackers to mark any on-hold COD order as paid using a valid Razorpay payment                                                                                                                            
  tuple from their own separate transaction.                                                                                                                                                                
                                                                                                                                                                                                            
  **Current behavior:**                                                                                                                                                                                     
  - Endpoint uses `checkAuthCredentials()` as permission_callback — a no-op                                                                                                                                 
    stub that always returns true. Anyone can call this endpoint without any                                                                                                                                
    authentication.                                                                                                                                                                                         
  - Endpoint accepts whatever `razorpay_order_id` the caller sends without                                                                                                                                  
    verifying it belongs to the specified WooCommerce order. An attacker can                                                                                                                                
    reuse a valid payment tuple from their own ₹1 purchase to mark any victim's                                                                                                                             
    COD order as paid.                                                                                                                                                                                      
                                                                                                                                                                                                            
  **New behavior:**                                                                                                                                                                                         
  - Endpoint requires a valid `X-Razorpay-Signature` HMAC header               
    (`checkHmacSignature`). Only Razorpay's backend (MCS) can call this endpoint.                                                                                                                           
  - The Razorpay order ID is retrieved from WooCommerce order meta (stored at                                                                                                                               
    order creation time) and validated against the client-supplied value before                                                                                                                             
    any processing. Cross-order reuse is rejected.                                                                                                                                                          
  - Signature is verified using the stored order ID, not the client-supplied one.                                                                                                                           
                                                                                                                                                                                                

  Changes

  includes/api/api.php

  Switched permission_callback for the prepay-COD endpoint from
  checkAuthCredentials (no-op) to checkHmacSignature (HMAC-SHA256
  validation using the merchant's stored signing secret).

  includes/api/prepay-cod.php

  Restructured the function with the following fixes:

  1. Moved wc_get_order() before signature verification — order must exist
  and be validated before any crypto operations.
  2. Fixed meta key construction — now correctly builds the session key based
  on whether the order is a Magic Checkout order:
  $is1ccOrder = $order->get_meta('is_magic_checkout_order');
  $sessionKey = ($is1ccOrder === 'yes')
      ? 'razorpay_order_id_1cc' . $orderId
      : 'razorpay_order_id' . $orderId;
  3. Added empty stored order ID check — rejects with razorpay_order_not_found
  if no Razorpay order was ever stored for this WC order.
  4. Signature now uses stored order ID — verifyPaymentSignature() uses
  $storedRazorpayOrderId (fetched from DB) instead of the client-supplied
  $razorpayOrderId, fully closing the IDOR window.

  woo-razorpay.php + readme.txt

  Version bump: 4.8.2 → 4.8.3

### DIFF
--- a/includes/api/prepay-cod.php
+++ b/includes/api/prepay-cod.php
@@ -12,13 +12,57 @@ function prepayCODOrder(WP_REST_Request $request): WP_REST_Response {
         $payload = $request->get_params();
         $wcRazorpay = new WC_Razorpay(false);
 
-        $orderId = $payload[WC_ORDER_ID];
+        $orderId          = $payload[WC_ORDER_ID];
         $razorpayPaymentId = $payload[RAZORPAY_PAYMENT_ID];
-        $razorpayOrderId = $payload[RAZORPAY_ORDER_ID];
-        $signature = $payload[RAZORPAY_SIGNATURE];
-        $razorpayOffer = $payload[RAZORPAY_OFFER];
+        $razorpayOrderId  = $payload[RAZORPAY_ORDER_ID];
+        $signature        = $payload[RAZORPAY_SIGNATURE];
+        $razorpayOffer    = $payload[RAZORPAY_OFFER];
+
+        // Validate the order exists before any further processing
+        $order = wc_get_order($orderId);
+        if (!$order)
+        {
+            return new WP_REST_Response(['code' => 'invalid_order'], 400);
+        }
+
+        // Prevent double-payment on an already-paid order
+        if ($order->is_paid())
+        {
+            return new WP_REST_Response(['code' => 'order_already_paid'], 400);
+        }
+
+        // Validate that the Razorpay order ID in the request matches what was stored
+        // when the WooCommerce order was created. The stored meta key depends on
+        // whether this is a 1CC (Magic Checkout) order or a standard checkout order:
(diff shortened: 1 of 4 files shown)


includes/api/prepay-cod.php
ramth05: @abdulwahidsharief @gutti-yaswanth , is this the correct way to verify the signature?

ramth05 (changes requested):

ChetanGN (approved):

ramth05 (approved):
LGTM

PR #665, opened 2026-08-26
[Security] Add HMAC signing to WooCommerce shipping-info call (PSID 1b9300d255b0)

abhilash-rp:
## Summary

Fixes a publicly disclosed unauthenticated IDOR (Patchstack PSID `1b9300d255b0`, CVSS 5.3, published 2026-08-17) in `POST /wp-json/1cc/v1/shipping/shipping-info`. Any unauthenticated caller could supply an arbitrary WooCommerce `order_id` and have the endpoint write shipping metadata onto it.

Follows the exact same two-part pattern as the already-merged **PR #647** (CVE-2026-3507, `cod/order/prepay`):

1. **Auth: HMAC signature required.** `includes/api/api.php` — the `shipping-info` route's `permission_callback` switches from the no-op `checkAuthCredentials` to `checkHmacSignature`. One line, exactly as #647 did it. **No API functionality changes** — no new/renamed parameters, no altered success responses, no contract changes.
2. **Authorization: order-ownership validation.** `includes/api/shipping-info.php` — looks up the target order, derives its stored Razorpay order ID (same `getOrderSessionKey()` pattern #647 used), and rejects with 400 unless the caller-supplied `razorpay_order_id` matches. Runs before any cart/state mutation, and the stored value (never the caller's) is used downstream.

Supporting changes:
- `includes/api/auth.php` — `checkHmacSignature()` now prefers `$request->get_body()` over `php://input` (makes it unit-testable) and logs a distinct code per rejection path (was previously silent on all three).
- `includes/utils.php` — the shipping validator now requires `razorpay_order_id`.
- `includes/api/shipping-info.php` — also fixes `add_post_meta` → `update_post_meta`, which let repeated calls stack duplicate `1cc_shippinginfo` rows (and caused a real HPOS/legacy divergence in the shipping line items written at order-update time).
- `tests/phpunit/tests/test-shipping-info-auth.php` — new, 12 cases: 3 auth-layer, 4 ownership-layer (incl. the IDOR regression test, which asserts the victim order's metadata is genuinely unmutated, not just a status code), 1 validation, 4 happy-path/idempotency.
- Version bump 4.8.7 → 4.8.8, changelog entry, `json_encode` → `wp_json_encode` in the new branches.

## ⚠️ REQUIRED DEPLOYMENT ORDER

Same constraint #647 and MCS PR #2499 documented, for the same reason:

1. **Deploy [magic-checkout-service#3692](https://github.com/razorpay/magic-checkout-service/pull/3692) first** — it makes MCS sign outbound shipping-options requests with `X-Razorpay-Signature`.
2. **Release this plugin (v4.8.8) after**, once signing is verified live in production.

**Reversing the order returns 403 on every merchant's shipping-options call until MCS catches up.** MCS #3692 also widens its fail-open branch to cover 401/403, which caps the blast radius of a partial rollout at degraded shipping rates rather than a dead checkout — but that is a safety net, not a substitute for the ordering.

Also confirm per-merchant signing-secret coverage (`platform_signing_secrets["woocommerce"]`) before release: a merchant without a secret gets no signature from MCS and will 403 here.

## Test plan
- [x] `./vendor/bin/phpunit` — **157 tests, 455 assertions, 0 errors, 0 failures, 0 skipped**
- [x] All 12 cases in `test-shipping-info-auth.php` pass, including the 3 auth-layer tests that verify HMAC rejection (missing / invalid / unconfigured-secret → 403)
- [x] IDOR regression test asserts victim-order metadata is unmutated
- [x] Manual: reproduced the IDOR pre-fix, confirmed it now returns `razorpay_order_id_mismatch` / 400
- [x] Verified only `shipping-info`'s callback changed — the six browser-called routes (`order/create`, `abandoned-cart`, `cart/fetch-cart`, `cart/create-cart`, `coupon/apply`, `giftcard/apply`) correctly remain on `checkAuthCredentials`, since the storefront JS cannot sign
- [ ] Plugin Check — approximated via a scoped PHPCS+WPCS ruleset (no standalone Plugin Check package on Packagist); 3 in-scope findings fixed. 189 pre-existing findings elsewhere in `woo-razorpay.php` flagged for separate follow-up, out of scope here

## Note for anyone running the test suite locally

`tests/phpunit/bootstrap.php` derives the active-plugin slug from `basename(realpath(...))` — the checkout directory name — while `vendor/valu/wp-testing-tools`' `wp-install` symlinks the plugin into `wp-content/plugins/` using `composer.json`'s package name (`razorpay-woocommerce`). **If your checkout directory isn't named exactly `razorpay-woocommerce`, the plugin never activates**, `WC_Razorpay` is undefined, and you'll see dozens of spurious `Class 'WC_Razorpay' not found` errors that look like real failures. Worth fixing separately in `bootstrap.php`.

## Follow-ups (deliberately not in this PR)
- **Telemetry on the rejection branches.** They log via `rzpLogError()`, which is gated on `enable_1cc_debug_mode` (defaults `no`) — so on most installs a blocked exploitation attempt currently produces no visible signal. Worth adding a `rzpTrackDataLake` event.
- **Monitor `razorpay_order_id_mismatch` rate post-release.** One narrow legitimate-mismatch path exists: if `createOrGetRazorpayOrderId()` regenerates the Razorpay order (on `verifyOrderAmount` failure) after MCS already holds the previous ID.
- **Extract the session-key derivation.** Now hand-copied in three places, two of which are security controls (`prepay-cod.php`, `shipping-info.php`) plus the canonical `WC_Razorpay::getOrderSessionKey()`.

## References
- Patchstack PSID `1b9300d255b0` (published 2026-08-17)
- Precedent: PR #647 (CVE-2026-3507, `prepayCODOrder` IDOR) and MCS PR #2499
- Companion: [magic-checkout-service#3692](https://github.com/razorpay/magic-checkout-service/pull/3692)

### DIFF
--- a/includes/api/api.php
+++ b/includes/api/api.php
@@ -77,7 +77,7 @@ function rzp1ccInitRestApi()
         array(
             'methods'             => 'POST',
             'callback'            => 'calculateShipping1cc',
-            'permission_callback' => 'checkAuthCredentials',
+            'permission_callback' => 'checkHmacSignature',
         )
     );
 
--- a/includes/api/auth.php
+++ b/includes/api/auth.php
@@ -29,16 +29,29 @@ function checkHmacSignature($request)
 
 	if (empty($signature))
 	{
+		rzpLogError('1cc_hmac_auth_failed: SIGNATURE_MISSING');
 		return new WP_Error('rest_forbidden', __('Signature missing'), array('status' => 403));
 	}
 
-    $payload = file_get_contents('php://input');
+	// Prefer the body WordPress already parsed; fall back to the raw stream.
+	// Both are the same bytes in production, but get_body() is populated under
+	// PHPUnit where php://input is empty, which makes this function testable.
+	$payload = '';
+	if ($request instanceof WP_REST_Request)
+	{
+		$payload = $request->get_body();
+	}
+	if (empty($payload))
+	{
+		$payload = file_get_contents('php://input');
+	}
 
 	// Retrieve 1CC signing HMAC secret saved at plugin load time
-    $secret = get_option('rzp1cc_hmac_secret');
+	$secret = get_option('rzp1cc_hmac_secret');
 
 	if (empty($secret))
 	{
+		rzpLogError('1cc_hmac_auth_failed: SECRET_NOT_CONFIGURED');
 		return new WP_Error('rest_forbidden', __('Secret not configured'), array('status' => 403));
 	}
 
@@ -51,6 +64,7 @@ function checkHmacSignature($request)
 	}
 	catch (Errors\SignatureVerificationError $e)
 	{
+		rzpLogError('1cc_hmac_auth_failed: SIGNATURE_INVALID');
 		return new WP_Error('rest_forbidden', __('Invalid signature'), array('status' => 403));
 	}
 
--- a/includes/api/shipping-info.php
+++ b/includes/api/shipping-info.php
@@ -44,6 +44,60 @@ function calculateShipping1cc(WP_REST_Request $request)
         $addresses    = $params['addresses'];
         $rzpOrderId   = sanitize_text_field($params['razorpay_order_id']);
 
+        // Bind the request to the target order before touching any state.
+        // Without this, any caller can name an arbitrary order id and have its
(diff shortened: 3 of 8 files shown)


ramth05 (approved):
lgtm

